PR #826, opened 2013-10-31
Added libsecret

ManiacTwister:
As mentioned in the other pull request ( #825 ) libsecret is the replacement for gnome-keyring, so here is the new implementation.

cheers

### DIFF
--- a/src/common/secret_schema.c
+++ b/src/common/secret_schema.c
@@ -1,0 +1,32 @@
+/* HexChat
+ * Copyright (C) 2013 Robin Rieger
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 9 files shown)


src/common/secret_schema.c
```diff
@@ -0,0 +1,32 @@
+/* HexChat
+ * Copyright (C) 2013 Robin Rieger
```
leeter: odd I didn't think this pull was that old... perhaps the copyright should be updated

ManiacTwister:
Happy 3. Birthday to my pullrequest

TingPing:
@ManiacTwister So what I really want to happen is some abstract interface for storing passwords with backends for OSX, Windows, libsecret, plaintext, etc. This PR, while perhaps functional, is sort of an ugly mess of ifdefs throughout the code base that is not easily extended. Also I don't think you ever handled migration?

vgturtle127:
@ManiacTwister @TingPing This seems like a cool idea, has anybody looked at it recently?

PR #904, opened 2014-02-27
Some extra gTLDs added.

sacarasc:
I will not swear about how much I dislike ICANN for allowing all these extra gTLDs, but this is part one.


TingPing:
I'll be honest, these don't interest me. The previous ones made sense because they were remotely related to location, they no longer are. Sorry.

Mikaela:
Where is that gTLD list actually used? Cannot HexChat just accept them all as long as they resolve somewhere?

There are other TLDs than what ICANN has and they seem to be coming all the time and there are also alternative roots such as OpenNNIC and there used to be 42registry.

sacarasc:
They're used in the right click user menu and the /country command.

Mikaela:
And how does that work with vhosts or cloaked IPs?

That aso makes no sense to me, because my reverse DNS can be anything and I don't necressarily have to be in that country in order to have the reverse DNS.

sacarasc:
It's a rough guide and nothing more. To be honest, that part of it is less useful than /country ever was.

PR #1162, opened 2014-09-30
Optimized and improved logo SVGs and PNGs.

obskyr:
I have optimized the SVGs - they no longer use more than 15kB of redundant data each, and are now all friendly paths and therefore renderable in browsers. The shadow of the Hexchat logo is also no longer cut off to the right. Tray icons, program icons, and miscellaneous PNGs have all been updated to reflect these changes. 

`hexchat-b.svg` doesn't seem to be used anywhere and should probably be removed, but I am not going to make that call.

I almost wanted to remove the RDF data just for size reasons, but I doubt people would like that very much - one thing I noticed and am wondering though - does Peter Zelezny really own the rights to the Hexchat logo?


obskyr:
I'd like to draw attention to this pull request once again, almost a year later. The only visual changes it makes is a small change to the tray icon's symbols (which was inevitable, as no SVGs were available for those) and removing the shadow on those (since the default tray icon, symbolless, was already shadowless). Beyond that, it's all just optimization and objective improvement.

Is there any reason not to merge this pull request at this point?

TingPing:
@obskyr Could you make the svg the same size as previously, it looks bad being smaller than everything else in launchers.

obskyr:
I'm assuming you mean remove the margin on `hexchat.svg` (and not simply just scaling up everything - if that's what you meant, that's no problem at all, just say the word). And in that case, sure! Problem there, though, is that the shadow will be cut off again and look pretty dumb. Basically, there are four ways to proceed here:

* Leave it as it is, with the shadow completely displayed but the actual logo part smaller relative to the image
* Crop the SVG to just the logo with the shadow being abruptly cut off (which looks egh, as shown for example on Wikipedia)
* Crop the SVG to just the logo and remove the shadow
* Keep the SVG, which is the "official logo", with the shadow, and remake the `.ico` with the shadowless version. If the `.ico` is what's used in launchers, this'll look way better.

Which one would you like me to do?

TingPing:
On Linux the svg is directly used so that is where the problem is. Of those I would be fine with either removing the shadow or perhaps redesigning the shadow to not be cut off at the same size.

obskyr:
Fixed you up three different versions. Check it:

* [Shadowless](http://obskyr.io/filer/hexchat-shadowless.svg) (literally just `hexchat-shadowless.svg`)
* [Smaller shadow](http://obskyr.io/filer/hexchat-smallershadow.svg)
* [Solid shadow](http://obskyr.io/filer/hexchat-solidshadow.svg)

Haven't optimized them yet, so just tell me which one you'd like and I'll optimize it, commit and PR.

TingPing:
@obskyr solid shadow looks good

PR #1205, opened 2014-11-09
Add Volatile to default networks list

wowaname:
Website at http://volatile.club/ with server list and webchat. Allows Tor and proxies, Unicode nicknames, and #&+ prefixed channels. Will eventually have a Python-powered services bot to replace what we have now.


wowaname:
Domain does not resolve anymore.

TingPing:
Good thing I never merge these right away :P

wowaname:
Haha, I have a new (permanent) domain and I may resubmit sometime, but I don't have much of an incentive to do so :p

PR #1257, opened 2015-01-02
Added device iteration to Sysinfo

KingHual:
Sysinfo now displays multiple iterations of the query results (specifically, VGA info)

### DIFF
--- a/plugins/sysinfo/sysinfo.cpp
+++ b/plugins/sysinfo/sysinfo.cpp
@@ -231,7 +231,7 @@ getMemoryInfo (void)
 static char *
 getWmiInfo (int mode)
 {
-	/* for more details about this wonderful API, see 
+	/* for more details about this wonderful API, see
 	http://msdn.microsoft.com/en-us/site/aa394138
 	http://msdn.microsoft.com/en-us/site/aa390423
 	http://msdn.microsoft.com/en-us/library/windows/desktop/aa394138%28v=vs.85%29.aspx
@@ -239,6 +239,7 @@ getWmiInfo (int mode)
 	*/
 
 	char *buffer = (char *) malloc (128);
+	char *buffer_iter = buffer;
 	HRESULT hres;
 	HRESULT hr;
 	IWbemLocator *pLoc = NULL;
@@ -323,6 +324,7 @@ getWmiInfo (int mode)
 		hr = pEnumerator->Next (WBEM_INFINITE, 1, &pclsObj, &uReturn);
 		if (0 == uReturn)
 		{
+			*(buffer_iter - 2) = 0;
 			break;
 		}
 		VARIANT vtProp;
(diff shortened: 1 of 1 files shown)


plugins/sysinfo/sysinfo.cpp
```diff
@@ -338,7 +340,10 @@ getWmiInfo (int mode)
 				hr = pclsObj->Get (L"Name", 0, &vtProp, 0, 0);
 				break;
 		}
-		WideCharToMultiByte (CP_ACP, 0, vtProp.bstrVal, -1, buffer, SysStringLen (vtProp.bstrVal)+1, NULL, NULL);
```
TingPing: While you are at it maybe fix the possible overflows.

KingHual: Alright.

plugins/sysinfo/sysinfo.cpp
```diff
@@ -338,7 +340,10 @@ getWmiInfo (int mode)
...
+		buffer_iter += WideCharToMultiByte(CP_ACP, 0, vtProp.bstrVal, -1, buffer_iter, SysStringLen(vtProp.bstrVal) + 1, NULL, NULL) - 1;
+		*(buffer_iter++) = ',';
```
TingPing: I haven't ran this or anything, but isn't adding "," at the end everytime wrong? It should be added at the start of a second loop if you have more than one?

KingHual: I wouldn't call it "wrong". I was thinking whether I should add one before or after the WideCharToMultiByte call, but I decided to go with after and then remove the last 2 chars after all iterations have been processed.

KingHual: Here's what it outputs for me:

* King_Hual ** SysInfo ** Client: HexChat 2.10.2 (x64) ** OS: Microsoft Windows 7 Home Premium  ** CPU: Intel(R) Core(TM) i3 CPU       M 350  @ 2.27GHz (2.00 GHz) ** RAM: 3893 MB Total (1316 MB Free) ** VGA: AMD Radeon HD 6500M/5600/5700 Series, Intel(R) HD Graphics ** Uptime: 11.27 Hours **

TingPing: >and then remove the last 2 chars after all iterations have been processed.

OH, I see that now. I suppose thats fine.

plugins/sysinfo/sysinfo.cpp
Arnavion: Using `WideCharToMultiByte` here is wrong anyway. It's passed to hexchat_commandf which expects utf-8. Use g_utf16_to_utf8

Also use a dynamic string instead of the fixed 128B buffer with g_strjoin. This will also save you the effort of deciding whether to append ", " or not.

plugins/sysinfo/sysinfo.cpp
```diff
@@ -338,7 +342,9 @@ getWmiInfo (int mode)
 				hr = pclsObj->Get (L"Name", 0, &vtProp, 0, 0);
 				break;
 		}
-		WideCharToMultiByte (CP_ACP, 0, vtProp.bstrVal, -1, buffer, SysStringLen (vtProp.bstrVal)+1, NULL, NULL);
+		buffer_iter += WideCharToMultiByte(CP_ACP, 0, vtProp.bstrVal, -1, buffer_iter, WMI_INFO_BUFFER_LENGTH - (buffer_iter - buffer + 2), NULL, NULL) - 1;
```
KingHual: Alright, I resolved the buffer overflow and added made buffer null-initialized to avoid any string function issues.

Arnavion:
https://github.com/hexchat/hexchat/commit/e758da5 fixed this and a lot of other things

PR #1263, opened 2015-01-06
Changed cps limit system from bytes to a more dynamic string system that allows for things like 1k 4G 3m and so on

WIsOH:


Arnavion:
s/GUI/UI/

That is to say, I will also accept allowing human-readable sizes via `/SET`.

Changing the persistence of that value in the config file and the internal structure that holds these values (prefs.hex_dcc_global_max_send_cps, etc.) to that human-readable string is utter madness.

PR #1273, opened 2015-01-19
servlist: Remove DeltaAnime (cannot be connected to)

CraigyDavi:


TingPing:
I can connect to it fine.

PR #1287, opened 2015-02-06
send +R instead of +r for "registered only"

devinacker:
HexChat currently sends lowercase +r ("channel is registered", which users can't change) instead of +R (the de facto "allow registered users only" mode) when pressing the "Registered Only" button on the topic bar. It seems like this might have been an oversight made when changing the buttons from uppercase to lowercase *(Arnavion edit: https://github.com/hexchat/hexchat/commit/7ba2f1f)*, but this is what the intended behavior was, as far as I can tell...

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -666,7 +666,7 @@ process_numeric (session * sess, int n,
 			EMIT_SIGNAL_TIMESTAMP (XP_TE_CHANMODES, sess, word[4], word_eol[5],
 										  NULL, NULL, 0, tags_data->timestamp);
 		fe_update_mode_buttons (sess, 'c', '-');
-		fe_update_mode_buttons (sess, 'r', '-');
+		fe_update_mode_buttons (sess, 'R', '-');
 		fe_update_mode_buttons (sess, 't', '-');
 		fe_update_mode_buttons (sess, 'n', '-');
 		fe_update_mode_buttons (sess, 'i', '-');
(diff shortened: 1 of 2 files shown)


TingPing:
+R does not work on freenode and +r does.https://freenode.net/using_the_network.shtml

devinacker:
Interesting, I didn't realize charybdis (and ircd-seven) differed in that aspect...

From what I can tell there aren't any major ircds that use +r for that purpose and also have capital-R in their CHANMODES; it'd probably work to just have a button for each and just hide one or the other depending on whether +R is available or not. Thoughts?

Mikaela:
Charybdis is quite major IRCd if you ask me. I am also unable to see CHANMODES including +R, but thta was just quick glance and I will be parsing those lists...

TingPing:
It should already hide modes that do not exist.

devinacker:
It does, but if an alternate button is added for networks that use +R for registered-only channels, the +r button would still be there since (almost?) all of them use that for a different mode as well.

TingPing:
Yea if they have both it would be duplicated and wrong. It is sort of hard to handle this in any correct way no matter what we have to make assumptions. Honestly we could just remove it, when I added it I didn't realize nobody could agree on the case for registered-only...

devinacker:
Fine with me, it doesn't seem like it's usually used often enough for a button to be that necessary anyway.

Kcchouette:
According to [RFC2812](http://tools.ietf.org/html/rfc2812), the flag must be "+r".

Mikaela:
> According to RFC2812, the flag must be "+r".

+r is entirely different flag that doesn't work outside of IRCnet and it marks your connection as restricted and I think that was prevents joining/parting channels or something other bad until you reconnect. I was tricked into setting it years ago...

The RFC doesn't even recognize registered users at IRC as Services aren't part of it.

PR #1293, opened 2015-02-11
Selection:  Unhilight old selection on keypress.

RichardHitt:
Fixes issue 943.

### DIFF
--- a/src/fe-gtk/xtext.c
+++ b/src/fe-gtk/xtext.c
@@ -1872,6 +1872,9 @@ gtk_xtext_unselect (GtkXText *xtext)
 {
 	xtext_buffer *buf = xtext->buffer;
 
+	if (buf->last_ent_start == NULL)
+		return;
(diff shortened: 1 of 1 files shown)


RichardHitt:
Detailed changes:
   * gtk_xtext_unselect()
      -- return if no buf->last_ent_start
      -- set xtext->mark_stamp to FALSE (from button_release)
   * gtk_xtext_button_release()
      -- If null selection, do nothing (now done in button_press)
   * gtk_xtext_button_press()
      -- use gtk_xtext_unselect() vs gtk_xtext_selection_clear()
   * gtk_xtext_selection_kill()
      -- remove conditionality on not-Windows.  Remove test.

In particular, the gtk_xtext_selection_kill() fix, in Windows with a detached channel and an attached channel, selection in one clears selection in the other.  Perhaps this relates to the 943 issue.

PR #1297, opened 2015-02-21
Welcome to 2015!

James-TR:
Updated the copyrights to 2015 since it's almost the end of February already.


TingPing:
Berke Viktor is no longer really involved in the project so we should prob change that to "The HexChat Team" or something vague.. I'll get around to it before 2016 maybe.

James-TR:
Does the HexChat collective even have a name?

PR #1307, opened 2015-03-06
chug event loop after printing text

halfline:
Some irc proxies send a large amount of scrollback when first
connecting. During this time hexchat freezes up because the
I/O blocks the main loop from running.

As a workaround, this commit makes sure to explicitly chug
the main loop after each line is added.

This keeps things interactive when first connecting to znc.

### DIFF
--- a/src/fe-gtk/fe-gtk.c
+++ b/src/fe-gtk/fe-gtk.c
@@ -637,6 +637,8 @@ fe_print_text (struct session *sess, char *text, time_t stamp,
 			   gboolean no_activity)
 {
 	PrintTextRaw (sess->res->buffer, (unsigned char *)text, prefs.hex_text_indent, stamp);
+	while (gtk_events_pending ())
+	      gtk_main_iteration ();
 
 	if (!no_activity && !sess->new_data && sess != current_tab &&
 		sess->gui->is_tab && !sess->nick_said)


Arnavion:
This will also help with reading large scrollback files, so it's not a bad idea.

Arnavion:
The docs for gtk_main_iteration say that it will return TRUE if it handles a call to gtk_main_quit, which means that it handles calls to gtk_main_quit in the first place. But that means the quit cannot be honored until the loop is over, i.e., the full scrollback has been read (from disk or from the bouncer) and displayed.

- Is the quit still honored after the loop is over?
- Should we have the loop break in this case so that HC terminates sooner?

halfline:
Note, the ad-hoc loop typically only runs one iteration. It's not going to sit their for more than a few milliseconds.

The main_loop_quit will be honored next time the code returns to the main loop, but the delete-event that leads to the main_loop_quit will get processed right away. In fact there's an inherent danger here, that code you may expect to be running straight from the main loop will get run from the print text handler. maybe a better approach then the first patch would be to make fe_print_text buffer what's supposed to go on screen and send it their incrementally from an idle handler, then we don't have to worry about reentrancy problems and the like.

Arnavion:
>but the delete-event that leads to the main_loop_quit will get processed right away. In fact there's an inherent danger here, that code you may expect to be running straight from the main loop will get run from the print text handler.

Correct, this what I was talking about. The handlers registered for application quit will run, but the application will not actually quit because the main loop will return in the middle of the printing code. That's why I asked if the code that runs the inner loop should check the return value to see if a quit was requested, and if so, terminate the application.

halfline:
I guess that's not sufficient to avoid the problem though.  

For one, you can't do an "orderly" shutdown from the middle of the print handler directly.  the only two options are "exit()" and "return".  That first will miss some shutdown logic and the second will require checking if the user is quitting from every caller of the print handler.

Also, you could imagine other badness like print text getting called from print text or something because the source to get more data from the server wakes up in the middle of the print text handler.

I think we basically need to reject the patch outright and come up with a different approach.

Arnavion:
>That first will miss some shutdown logic and the second will require checking if the user is quitting from every caller of the print handler.

No, there is nothing important that will be missed by the first. The only thing that's done after fe_main() and gtk_main() have returned is freeing some resources, so calling exit() would be a viable strategy (and what I was originally thinking of). Having all the callers return a value whether the user requested a quit all the way up the callstack to the main event loop isn't really worth it.

>Also, you could imagine other badness

Oh yeah, that too.

PR #1317, opened 2015-03-14
Allow option to hide voice change messages

pR0Ps:
This patch adds an option to hide the "X gives voice to Y" and "X removes voices from Y" messages.

Screenshot of the addedoption:
screenshot of the added option

### DIFF
--- a/src/common/cfgfiles.c
+++ b/src/common/cfgfiles.c
@@ -502,6 +502,7 @@ const struct prefs vars[] =
 	{"irc_conf_mode", P_OFFINT (hex_irc_conf_mode), TYPE_BOOL},
 	{"irc_extra_hilight", P_OFFSET (hex_irc_extra_hilight), TYPE_STR},
 	{"irc_hide_nickchange", P_OFFINT (hex_irc_hide_nickchange), TYPE_BOOL},
+	{"irc_hide_voicechange", P_OFFINT (hex_irc_hide_voicechange), TYPE_BOOL},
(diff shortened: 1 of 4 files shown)


Arnavion:
1. Nobody wants a whole config for something so specific.
1. Just set the corresponding text events to the empty string.

pR0Ps:
Nobody? Some IRC networks use voice to represent state (-v = away, +v = back) and the voice changes can become annoying (full pages of just voice changes). See [here](https://robots.thoughtbot.com/weechat-for-slacks-irc-gateway#TSITfo) where somebody using WeeChat had the same problem and used the `hidevoices` filter to hide all the voice change messages. I realize this is the IRC network's problem and not Hexchats, but the ability to deal with it would be nice.

Would implementing this on a per-server basis be acceptable?

Is 2 a comment on how I should've implemented this or what I should do instead of submiting a patch?
I mimicked the way the [supression of nick change events](https://github.com/hexchat/hexchat/commit/0dccf692d41f62611b0b8c753ef4e84be670c736) was implemented on purpose since it's basically the same thing (mute a specific event).

Is this not the way to do it?

Arnavion:
>Nobody?

The keyword is "for something so specific." Why an option for suppressing voice changes? Why not one for halfops, one for ops, one for chanops... Just because you're using a network that overloads voice to represent away status doesn't mean we make an option tailored just for your use case.

>Is 2 a comment on how I should've implemented this or what I should do instead of submiting a patch?

The latter.

Farow:
You can easily make a script that hides voices per network or just set the text event to an empty string to disable it globally, like Arnavion said.

PR #1321, opened 2015-03-15
Text Events:  Allow any delimiter around nick, not just '<' and '>'

### LINKED ISSUE #1316: Using other characters besides greater/less than for messages breaks linking
Using things like this for "Channel Message" and "Your Channel" seems to break the linking (clicking their name directly in the main chat window).

`($1) $2`
`{$1} $2`

I know that the characters in the second example could be used in nicks, which is why I thought at first would be the problem, but not even () works.

RichardHitt:
FIxes #1316


TingPing:
Ok so this doesn't really fix it, it only makes it worse. Your changes:

- Don't treat '<' and '>' as special.
- Do treat '\b' (HIDDEN) as special.

So this fixes an event like this `%H(%Hnick%H)%H`
but breaks an event like this `<nick>`
and never fixed an event like this `(nick)`

PR #1327, opened 2015-03-21
Update ja_JP.po

iGlitch:
got busy couldn't finish, maybe i'll do the rest later


TingPing:
All translations are handled on transifex: https://www.transifex.com/projects/p/hexchat/

PR #1407, opened 2015-06-23
Add MadIRC to default networks list

Sheogorath-SI:
To check the FAQ "should" ideas:

`A maintained website listing servers.`
https://stats.madirc.net

`Has been around for a while.`
Founded on 4.6.2014

`A decent amount of active users.`
There are 30-50 Users around

`Doesn’t block large groups of users.`
Well, checkout by yourself :D


TingPing:

PR #1417, opened 2015-07-06
Windows 10 ID

ElPumpo:
Added windows 10 id.

### DIFF
--- a/win32/hexchat.exe.manifest
+++ b/win32/hexchat.exe.manifest
@@ -27,6 +27,13 @@
       <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
       <!--This Id value indicates the application supports Windows 8.1 functionality-->
       <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+      <!--This Id value indicates the application supports Windows 10 functionality-->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
     </application>
   </compatibility>
+    <application xmlns="urn:schemas-microsoft-com:asm.v3">
+    <windowsSettings>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+    </windowsSettings>
+  </application>
 </assembly>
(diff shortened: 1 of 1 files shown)


TingPing:
Do you have a hidpi monitor and does everything work perfectly for you? Another user had reported such but it doesn't really make sense. GTK2 is certainly not actually DPI aware on any platform unless it happens to 'just work' on win32 which my understanding was that wouldn't be true.

ElPumpo:
Yeah I got a hidpi monitor, and gtk2 doesn't have to be dpi aware. 
If you could compile a version that is dpi aware I would like to try it out.

TingPing:
> Yeah I got a hidpi monitor, and gtk2 doesn't have to be dpi aware. 

So everything in hexchat is super small for you then?

> If you could compile a version that is dpi aware I would like to try it out.

It isn't and never will be DPI aware.

ElPumpo:
No, not everything is super small.

The automatic Windows font scaler only works up to 100% size at a 1440p res on Windows display settings.
Indicating that the application is DPI aware it will by itself scale the font so Windows doesn't wash out the fonts by trying to make them look good.

ATM the application has tiny and washed out fonts.
I've used DPI aware on my personal applications with success.

If you could build a version that is DPI aware and send it to me so I can test the application myself on a hidpi monitor it would be great. I'll give you screenshots to compare too.

If you're never making your application DPI aware hidpi display users will have to look another direction.

TingPing:
> If you're never making your application DPI aware hidpi display users will have to look another direction.

That's not up to hexchat; Gtk2 is the toolkit used for the application and it is no longer actively getting new features such as hidpi support.

ElPumpo:
Oh, I understand. Is there a possibility to upgrade from Gtk2 to Gtk3 without chaning major code, as Gtk 3.10 added hidpi support?

DoctorWho11:
Regardless, there would need some changes as a lot of gtk2 api has been deprecated. And the xtext widget would need to be updated to make use of cairo drawing.

ElPumpo:
Yeah, upgrading from Gtk2 to Gtk3 isn't going to be a switch.

Anyways I found a fix:
1. Right click hexchat.exe and press properties
2. On the compatibility tab there is a "disable scaling at high DPI" (i'm on a Swedish OS, using sence-translate). Check the button and restart the client.

 Works great now! I've done this with a lot of applications like Steam and Spotify; works great!

But still official HidPI support would be better.

TingPing:
This was done in https://github.com/hexchat/hexchat/commit/e40fbee772f13abb125e47cd7a1525689756ae11

PR #1434, opened 2015-07-29
Correct a translation with the wrong meaning

palopezv:
In Spanish, "Download list" is "Descargar lista", not "Lista de descarga":  an itemization of things to download, not the intended meaning in the menu entry.


TingPing:
All translations are handled on transifex: https://www.transifex.com/projects/p/hexchat/

PR #1457, opened 2015-08-21
Be forgiving of a missing ending CTCP delimiter in a truncated message

arodland:
Some clients don't bother splitting messages to avoid exceeding the IRC line length limit, and some clients do, but they don't correctly split CTCP messages. In either case, we might receive a CTCP message from the server that's missing the ending 0x01 delimiter.

In that case, HexChat won't recognize it as a CTCP message and the result is something like this: 
hexchat-action

I think that there's very little harm in assuming that any message that begins with 0x01 is intended as CTCP, since there's no other legitimate use, and the code change required is very small.


arodland:
If you don't like this, a middle ground would be to actually check if the message is maximal-length and only forgive the missing delimiter then :)

Arnavion:
In your screenshot, is it that p5commit's message was too long and truncated by the server?

arodland:
Correct.

DanielOaks:
This'd be a good change if possible, ending delimiters can get dropped fairly easily if other clients aren't super careful.

PR #1464, opened 2015-09-01
Reload python script on /load if already loaded

rredford:
Add check if script was already running, if so, it just reuses reload code and returns.


TingPing:
- Doesn't actually build...
- What about Perl scripts?
- What about C plugins?

rredford:
Yeah thought it'd be straightforward, but I hope you understand the idea and make it work. :)


-------- Original Message --------
Subject: Re: [hexchat] add check if already running (#1464)

PR #1466, opened 2015-09-01
Automatically adjust the maximum indent size on HiDPI displays.

LawnGnome:
Currently, the indent size is always defined in pixels, which means that the default behaviour on a 168 dpi display when timestamps are enabled is that nicks will start overlapping with timestamps when they're seven or more characters long. This is unfortunate.

This change adjusts text_max_size on displays with 120 dpi or more to be relative to a hypothetical 96 dpi display, which results in better out of the box behaviour on my Retina MBP (on Linux, at least).

Potential issues with this PR:

* `gdk_window_get_screen()` requires GDK 2.24 or newer. I don't know what HexChat's minimum version requirement is. This could be polyfilled with the older `gdk_drawable_get_screen()` if support for older versions is needed.

* Getting the resolution each time `gtk_xtext_append_indent()` is called may be expensive, although I don't notice any jank. It could be cached in the GtkXText structure if necessary.

* I haven't tested this beyond my Ubuntu 15.04 box.

### DIFF
--- a/src/fe-gtk/xtext.c
+++ b/src/fe-gtk/xtext.c
@@ -4676,24 +4676,40 @@ gtk_xtext_append_indent (xtext_buffer *buf,
 		space = 0;
 
 	/* do we need to auto adjust the separator position? */
-	if (buf->xtext->auto_indent &&
-		 buf->indent < buf->xtext->max_auto_indent &&
-		 ent->indent < MARGIN + space)
-	{
-		tempindent = MARGIN + space + buf->xtext->space_width + left_width;
+	if (buf->xtext->auto_indent) {
+		int max_auto_indent = buf->xtext->max_auto_indent;
+		GdkScreen *screen = gdk_window_get_screen (buf->xtext->draw_buf);
+		double gdk_reso = gdk_screen_get_resolution (screen);
 
-		if (tempindent > buf->indent)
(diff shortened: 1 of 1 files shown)


TingPing:
>  I don't know what HexChat's minimum version requirement is. 

2.24 is required.

> Getting the resolution each time gtk_xtext_append_indent() is called may be expensive, although I don't notice any jank. It could be cached in the GtkXText structure if necessary.

Seems reasonable to cache, can hook the *resolution* property for changes also.

Ping @RichardHitt to review this

TingPing:
Some possible things to think about:

- Gtk2 itself isn't hidpi aware in anyway, so for one aspect of the GUI to be based upon DPI seems odd. To me personally it makes more sense to base indent size on font width.
- There is the question of variable DPI, for example what happens with two monitors running wayland at different DPIs and you drag hexchat between them.

Thumpxr:
For me on Windows 10 its still blurry (2.12.1).
Will this PR be implemented?

TingPing:
> For me on Windows 10 its still blurry (2.12.1).

This PR has nothing to do with that.

Anyway I think I will just be closing this with my previous comment really summing it up. Gtk2 isn't dpi aware and the only good solution to that is moving to Gtk3.

PR #1487, opened 2015-09-21
fix 'make dist' in appindicator

praiskup:


TingPing:
Fixes go into master.

PR #1492, opened 2015-09-29
Remove AMSG as a default macro

Xe:


berrythesoftwarecodeprogrammar:
NO

KindOne-:
Yes. We don't need an IRC client including ways to spam all the channels.

berrythesoftwarecodeprogrammar:
ihu

Mikaela:
:+1: I have never seen proper use case for that (`/allchan` and /`allserv` are useful) and people just start complaining that "I just greet all channels with /amsg, what is the problem with it?"

If you really must greet all channels, do it separately and follow the discussion. The person above we thought to be spambot as they never did anything else than greeted in whatever that language was (there was also complaining that they are allowed to use whatever language they wish and what does it matter if we don't understand them) around the same time for three days.

TingPing:
> Yes. We don't need an IRC client including ways to spam all the channels.

This only stops accidental spam which I think is pretty uncommon, it doesn't take a genius to figure out `allchan say hi`.

Mikaela:
While being uncommon it still happens. It's not so long time ago that I have last seen `/amsg what does this do?`

TingPing:
Well I don't think it is a particularly harmful command but I also don't think it is valuable, so unless somebody gives a use-case for it that seems useful I am fine with removing it. (note it will take *years* to even propagate this change to most users)

berrythesoftwarecodeprogrammar:
lol

erm:
this command is extremely valuable and important, do not remove it

TingPing:
> this command is extremely valuable and important, do not remove it

Do tell.

Mikaela:
> this command is extremely valuable and important, do not remove it

Could you explain how it's *extremely valuable and important*?

killkikes:
/amsg is not only useless but inflammatory and detrimental to the workflow of normal IRC function and only spammers and general anarchists want amsg to stay and for that alone it so should be removed.

killkikes:
ps berrythesoftwarecodeprogrammar and erm are spammers trying to get this to not be removed be sure to make sure they dont win!

berrythesoftwarecodeprogrammar:
says the guy with the anti-semitic troll name. you are spreading lies and oppressing those who use irc differently to you.

berrythesoftwarecodeprogrammar:
you want to remove a feature used by many which is present in every irc client that matters, just because you personally dont use it. that is disgusting if you ask me.

tigercommander:
smh anti semite trolls

Poorchop:
The use case is pretty simple whether or not you agree with it - people simply wish to message all channels. Some people might use this command maliciously, and others might use it without any ill intent.

Removing this command isn't going to stop spam, nor will it have any visible effect on the amount of spam produced. People who want to screw around are always going to find a way at the end of the day, and considering that scripting is a far more useful route, why not just delete the Python and Perl plugins?

You deleted two lines. Even a skiddie won't have any issues with spending two seconds to add them back in. For the record, this command is actually really useful on certain groups of networks.

TingPing:
@Poorchop I already said as much regarding spammers.

Anyway yea this really doesn't matter, new users finding `/amsg` is hardly a big deal and spammers will be spammers.

Arnavion:
As a general recommendation, if you want to remove a feature, try to do more than sending a PR with no comment. The lack of any comment implies there is no question it will be accepted, as if the content of the PR is obvious, which in the case of removing functionality is not so.

wowaname:
>Removing this command isn't going to stop spam, nor will it have any visible effect on the amount of spam produced. People who want to screw around are always going to find a way at the end of the day, and considering that scripting is a far more useful route, why not just delete the Python and Perl plugins?

@Poorchop presents a perfect point. What's the reason for removing something that can be added back anyway?

Arguably acceptable uses for AMSG? Here:
* ``/amsg happy new year`` (or some other occasion where if someone gets mad that you're doing an amsg, they probably have a stick up their ass anyway)
* ``/amsg global announcement: (insert something serious here that everyone who knows the person should be aware of)``

Keep in mind that not all people are in five / fifty / five hundred channels across three / thirty / three hundred networks and they can get away with using /amsg in the channels they're on. Just because it is commonly seen as a nuisance doesn't translate into "it's useless, let's remove it for everyone, yada yada yada". With that logic, let's get rid of bold/underline/colour/reverse as well...

killkikes:
get cucked

Mikaela:
> /amsg happy new year (or some other occasion where if someone gets mad that you're doing an amsg, they probably have a stick up their ass anyway)

Pidgin had a nice feature to do this automatically and it got me quieted on #freenode some years ago.

> /amsg global announcement: (insert something serious here that everyone who knows the person should be aware of)

I think different anti spam bots trigger to this as fake global notice or similar.

wowaname:
>> /amsg happy new year (or some other occasion where if someone gets mad that you're doing an amsg, they probably have a stick up their ass anyway)
> 
> Pidgin had a nice feature to do this automatically and it got me quieted on #freenode some years ago.
Congrats. Not everyone is in #freenode though.

>> /amsg global announcement: (insert something serious here that everyone who knows the person should be aware of)
> 
> I think different anti spam bots trigger to this as fake global notice or similar.
Great. Not everyone is in those channels though.

tigercommander:
wowaname touches kids

PR #1504, opened 2015-10-07
Remove unnecessary information from the CTCP VERSION response.

SaberUK:
CTCP VERSION is for version information about the client not the system. There is no legitimate reason to expose information about the system in it.


TingPing:
I think there is value in knowing what OS you are on from both a support and stats perspective.

SaberUK:
Never mind then.

PR #1511, opened 2015-10-14
Added gorfnets network (fixes #1485)

### LINKED ISSUE #1485: add to network list: gorfnets (gorf.chat)
seems like this is a thing people do here so, could you add my network to the list?

name: gorfnets
address: gorf.chat
main ssl port: 6697
main clear port: 6667

other information available on https://gorf.chat/

ani6gup:


berrythesoftwarecodeprogrammar:
thanx fam

TingPing:

erm:
What is the criteria for including a network anyways?

TingPing:
@erm: https://hexchat.readthedocs.org/en/latest/faq.html#how-do-i-add-a-network-to-the-official-list

erm:
@TingPing:
Your link claims that it isn't a list of requirements, rather it is a list of guidelines. It would seem gorf.chat satisfies at least three of the four guidelines and it has done so for many years. I took the time to connect to and research some of the other included networks and more than one of them failed to meet any number of these guidelines.

I don't understand why gorf.chat is being excluded.

TingPing:
> Current global users 47

It just seems a bit small. The default network list doesn't need to contain every network just ones that would commonly be looked for.

erm:
Then why isn't that in the guidelines? The documentation doesn't appear to define any particular purpose for the default network list, nor do I think the default network list is primarily used in the manner you have described.  To many users, the default network list serves as a guide of to many potential quality networks that exist for any number of purposes. The population size of a network does not necessarily reflect the quality of that network's users or culture. Gorfnets is a quality network that satisfies three of the four guidelines, is stable, has longevity, and in my opinion possesses one of the most unique and vibrant cultures on IRC today.

While I agree that it would be a cumbersome, futile task to include all IRC networks, why not include a network like this when it explicitly requests inclusion on the default network list and satisfies the majority of the guidelines in the official documentation?

erm:
If you aren't going to include the network, could you at least update the documentation to be less ambiguous and ensure that the other included networks are compliant with whatever standard you may arrive at? Otherwise, this seems like it is possibly a biased exclusion.

TingPing:
I told you exactly why its not included and the docs already say that having a decent amount of users is part of what I look for. At a later date we can re-evaluate it if the network grows but ranting about it here does you no favors (I presume you are affiliated with the network?).

erm:
I do not represent the network, nor do I possess any privileges there, I'm just a chatter on the network. I've seen the population fluctuate to upwards of 100+ active users over the years, but it generally has had what I would deem a "decent amount" of users.

PR #1513, opened 2015-10-16
build: Let 'make dist' do a better job

ricotz:
Fix some 'make dist' issues to create a working tarball.
Making 'make distcheck' work is another story though.


TingPing:
Yea its a long way to getting everything in `make dist` so I eventually started ignoring it. =\

PR #1525, opened 2015-10-27
Update .travis.yml

paladox:
Add sudo: true.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ notifications:
         on_success: change
 matrix:
      fast_finish: true
+
+sudo: true


paladox:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

Arnavion:
It's been the default since they introduced it months ago, though maybe being future-proof doesn't hurt.

paladox:
Hi I doint think it is default if you were with travis ci before January 2015 I think.

•For repos we recognize before 2015-01-01, linux builds are sent to our standard infrastructure. 
•For repos we recognize on or after 2015-01-01, linux builds are sent to our container-based infrastructure.

TingPing:
We have been using travis since 2013, clearly it still builds fine so it is the default for us at least.

paladox:
I mean tests run fast. It will run on there new server with more resources available.

TingPing:
Not sure what you mean, this commit explicitly keeps us on the old servers.

From reading the docs I think it should be fine for us to move to the new system entirely, they have a new 'addon' to install apt packages and we have no other use for sudo.

You want to look into doing that for us? =) Somebody just needs to ensure everything we need is in the [whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise)

paladox:
Woops sorry I thought it was sudo: false. But I am not sure how to convert the code to the new server code.

TingPing:
You linked how to: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade#Adding-APT-Packages

paladox:
Yes but not sure how to change this

before_script:
    - sudo apt-get update
    - sudo apt-get build-dep -qq xchat
    - sudo apt-get install -qq libnotify-dev libproxy-dev libpci-dev libcanberra-dev monodevelop intltool

since I doint thing xchat is a package. The code is creating the package.

TingPing:
#1527 supersedes this, thanks for bringing it to our attention though.

Issue #1509, opened 2015-10-12
Update the copyright year


paladox:
Please update the copyright year that is displayed on the about window please.

Also move the check for update and move it about the about menu. since about should be at the bottom of the menu and check for update should be in the middle.

TingPing:
The updater is an external plugin and has no control over the about window.

paladox:
I mean move it up places in the menu, keep it in the current menu but move its position so check update is above about.

paladox:
Sorry didn't read your comment correctly, I see you know what I mean thanks.

trishnaguha:
I would like to work on this issue. How to run the local instance of hexchat?
Thanks!

TingPing:
> How to run the local instance of hexchat?

As in you built it on linux? `./src/fe-gtk/hexchat` or just install it.

trishnaguha:
@TingPing Yes I am on fedora.

trishnaguha:
I am unable to find the check for update option in hexchat and does about menu here refer to about window?

paladox:
Thanks.

trishnaguha:
I have found the file for "Check for updates" https://github.com/hexchat/hexchat/blob/master/plugins/upd/upd.c
But I am not able to find this option "Check for updates"  on my irc client hexchat on linux. 
Thanks

TingPing:
@trishnaguha The updater is windows only, on linux thats up to your distro ofc.

trishnaguha:
@TingPing okay thanks

PR #1528, opened 2015-10-28
Copyright year updated

trishnaguha:
I have updated the copyright year according to the issue https://github.com/hexchat/hexchat/issues/1509


paladox:
Thanks.

Arnavion:
It's not really necessary... and if you really want it to be up-to-date for aesthetic reasons then wait another 2 months.

TingPing:
Also bviktor isn't really a part of the project any more. Not sure what value the copyright in the about page even has, there are probably over 100 people with copyright over part of the project.

PR #1531, opened 2015-11-01
Update translations

zygimantus:


TingPing:
All translations are handled on Transifex: https://www.transifex.com/hexchat/hexchat/

PR #1532, opened 2015-11-01
Get rid of scratch buffer in xtext entirely

mniip:
`gtk_xtext_strip_color` already supports allocation of a temporary buffer, why not use that?

On the downside, now the line length is unlimited, so this can lead to issues with scripts printing very large things. For example  `url_check_word` takes a few seconds on 5k of data.


TingPing:
Not a fan of allocating an immediately freeing just to get the length everywhere.

mniip:
Could make a function that does that, separately, then.

PR #1549, opened 2015-11-20
Correctly check SSL_get_verify_result()

jakkdu:
According to OpenSSL document (https://www.openssl.org/docs/manmaster/ssl/SSL_get_verify_result.html),
when using SSL_get_verify_result(), the existence of certificate
needs to be checked. However, in current code, it does not.
Therefore, certificate existence check is required for correctly
handling the exception.

### DIFF
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -570,8 +570,6 @@ ssl_do_connect (server * serv)
 		} else
 		{
 			g_snprintf (buf, sizeof (buf), " * No Certificate");
-			EMIT_SIGNAL (XP_TE_SSLMESSAGE, serv->server_session, buf, NULL, NULL,
-							 NULL, 0);
 			goto conn_fail;
 		}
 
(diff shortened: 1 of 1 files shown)


TingPing:
Looks like you need to set an error message in `buf`.

jakkdu:
It seems better to use existing code with better error handling.

src/common/server.c
```diff
@@ -572,6 +572,7 @@ ssl_do_connect (server * serv)
 			g_snprintf (buf, sizeof (buf), " * No Certificate");
 			EMIT_SIGNAL (XP_TE_SSLMESSAGE, serv->server_session, buf, NULL, NULL,
```
TingPing: Probably should remove this print then to avoid duplicate messages, also reform the above to fit in.

jakkdu:
Yes. I removed previous error message printing.

Issue #702, opened 2013-08-02
Clicking URL doesn't always fire browser


deckblad:
I must often click URLs multiple times before a callout to a browser is made.  I've heard of this issue happening with other people too.

Win7 x64 w/ HexChat x64 v2.9.5

Issue #363, opened 2013-01-06
URLs not loading all the time when gui_url_mod is 0


AlannahDaw:
Sometimes when left clicking an URL it does not load, even though gui_url_mod is set to 0. 

I thought it might have just been something that only I experienced, but today I heard more anecdotal evidence that it doesn't always work for others.

Steps to reproduce:
/set gui_url_mod 0
Click lots of links until one fails.

MightyPork:
It is not a bug [explains how it's broken and a possible workaround]

really? Why not admit it's a bug at least?

PR #1581, opened 2016-01-07
Fix pixel-perfect clicks to open links in chat

### LINKED ISSUE #1414: Clicking a link very frequently fails
Seems like at least 50% of the time, clicking a link in a channel does nothing at all.  Mostly I have to resort to right-clicking and choosing "Open Link in Browser" to get it to work.  There doesn't seem to be any rhyme or reason to which ones work and which don't, either.

This is v.2.10.2 portable on Win7.

xPaw:
Emit word_click even as long as user is not selecting any text when releasing button

This fixes #1414, #363, #702.

@TingPing tested this change and he said it works as expected.

### DIFF
--- a/src/fe-gtk/xtext.c
+++ b/src/fe-gtk/xtext.c
@@ -113,6 +113,7 @@ char *nocasestrstr (const char *text, const char *tofind);	/* util.c */
 int xtext_get_stamp_str (time_t, char **);
 static void gtk_xtext_render_page (GtkXText * xtext);
 static void gtk_xtext_calc_lines (xtext_buffer *buf, int);
+static gboolean gtk_xtext_is_selecting (GtkXText *xtext);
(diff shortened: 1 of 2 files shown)


Arnavion:
The `while (ent)` can apparently become an infinite loop. Happened to me on the first link I clicked to test this.

TingPing:
`ent == buf->last_ent_end` should probably break?

xPaw:
My bad, should be fixed now.

TingPing:
@xPaw Might as well just use a for loop.

Arnavion:
```
<TingPing> Arnavion, selecting and lifting on a url does open here
<Arnavion> Because his for-loop version is wrong
<Arnavion> It moves the == end check to the precondition of each loop iteration instead of post
```

Arnavion:
Pushed fixed version as 971e5cf . Thanks.

PR #1598, opened 2016-01-30
Update travis badge image link

paladox:


TingPing:
Why? Is img.shields.io just down?

paladox:
Not sure but it is happening to others https://github.com/badges/shields/issues/640

PR #1618, opened 2016-02-23
SSLv23_client_method() is deprecated -> TLS

basisbit:
SSLv23_client_method() is deprecated. Use TLS_client_method (support TLSv1.2, TLSv1.1 and TLSv1) instead. see https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html


basisbit:
update to OpenSSL 1.1 and it will work.

basisbit:
never mind... OpenSSL documentation is ahead of current state of actual published stable OpenSSL versions. Reapply this after switching to OpenSSL 1.1

PR #1631, opened 2016-03-13
PR 1320

Arnavion:


vgturtle127:
@Arnavion @pstratem So what does this do? The title isn't very informative and the code changes are a little confusing.

This obviously affects the calculation of the lag, but what does that fix in itself?

Also, does timeout have to occur only when the lag is bad? It also occurs after long periods of inactivity, doesn't it?

Arnavion:
I've added the link to original discussion.

vgturtle127:
@Arnavion OK, thanks. I didn't know about the link, sorry.

Issue #1640, opened 2016-03-14
mIRC colors should actually match mIRC


someperson:
HexChat's "mIRC Colors" by default do not match the actual colors used by mIRC. This can cause problems with some color combinations, which are easily readable with mIRC's palette, but much more difficult to read with HexChat's default settings.

Since HexChat is supporting mIRC's standard here, I think it's reasonable for it to implement the colors defined in that standard, which are documented [here](http://www.mirc.com/colors.html).

![palettes](https://cloud.githubusercontent.com/assets/376450/13733800/4853fd30-e965-11e5-869c-b66cbc721422.png)

Arnavion:
>standard

That word doesn't mean what you think it means.

If you want the colors to be precisely equal to mIRC's then use the mIRC theme. Even the "standard" document you linked mentions that clients are free to have different colors.

HC's default colors seem to have a general muted theme to them, and I'm open to giving them slightly more contrast. But your justification ("must match mIRC") is invalid and your examples (blue on blue, red on red, and dark pink on dark teal) are unconvincing.

someperson:
mIRC is the de facto standard for colors on IRC. It is telling that the setting in HexChat is literally called "mIRC colors". I am not arguing that the colors must be exactly identical to the ones used in mIRC, but I think they should at least behave similarly. Namely, if a combination of colors results in readable text on mIRC, it should be equally readable in HexChat. The simplest way to ensure that text always displays as intended is to simply match the colors suggested in the standard.

Also note that I am not suggesting changing any colors used for the interface. This makes switching to the mIRC theme not a proper solution to this problem, since this changes far more than just the mIRC colors.

Arnavion:
>mIRC is the de facto standard for colors on IRC

> It is telling that the setting in HexChat is literally called "mIRC colors"

Again, mIRC does not get to decide the RGB values of each color, and no other client is under some obligation to use the same values as it.

>I am not arguing that the colors must be exactly identical to the ones used in mIRC

![image](https://cloud.githubusercontent.com/assets/1096010/13734944/a88e917e-e960-11e5-8f3c-22cf4259524a.png)

![image](https://cloud.githubusercontent.com/assets/1096010/13734948/b9576d32-e960-11e5-96f1-bbd8649d42e5.png)

>but I think they should at least behave similarly

Which they already do. Red is red. Green is green.

>Namely, if a combination of colors results in readable text on mIRC,

Which your examples aren't, as I already said.

---

[...]

PR #1641, opened 2016-03-14
Changed mIRC colors to match mIRC

someperson:
This is an implementation of #1640


TingPing:
As mentioned the docs say: 

> Other IRC clients may use slightly different colors for the above indexes

So I have no interest in exactly matching those colors in the default theme.

PR #1661, opened 2016-03-27
Typo fix.

sacarasc:
lintian told sney who told me to do it.

### DIFF
--- a/src/common/textevents.in
+++ b/src/common/textevents.in
@@ -304,7 +304,7 @@ pevt_generic_nick_help
 %C24*%O$tAlready offering CHAT to %C18$1%O
 1
 
-DCC Conection Failed
+DCC Connection Failed
 XP_TE_DCCCONFAIL
 pevt_dccconfail_help
 %C20*%O$tDCC $1 connect attempt to %C18$2%O failed (%C20$3%O)


TingPing:
Text events cannot be renamed because they will break every script and every theme using them.

vgturtle127:
@TingPing So if a script or theme uses "conection" instead of "connection" and we change it, you are saying it would break all scripts and themes that use that? It would be an extremely simple fix, though, right? Just spell it correctly?

TingPing:
Yes changing it would break everything that used the old spelling. I agree it is a minor change but I'd rather not break API for a minor change.

vgturtle127:
@TingPing But wouldn't any theme or script be super easy to update by just using find and replace and changing all instances of "Conection" to "Connection"? And all themes and scripts are easily editable, right?

I'm just saying having an announcement that there were some typos in the source that were updated and developers of themes and scripts need to update them (simply by using the correct spelling) doesn't seem to be that big of deal, at least in my eyes. I can understand not wanting to switch an entire codebase from one external library to another, or from one API to another, but this is really just a small enough change that I think the developers wouldn't mind it. Especially since most of them probably have noticed the incorrect spelling before.

Is there a repository or something of scripts and themes?

Arnavion:
It's absurd to break existing addons and scripts just to fix a typo. The typo is public API now.

Any talk of "surely script authors will notice the event name changed and update their scripts" ignores reality. There are far more addons than in our repository; script authors may or may not update their scripts; users will almost certainly not update their scripts in step with the new HC update and so blame HC for breaking their scripts; and even users who know how to edit scripts will be stumped by compiled addons that don't provide source.

vgturtle127:
@Arnavion That's all fair enough, but when was the last time the API changed? If it has been a long time, then I totally understand. Although, if the API is that old, maybe it needs to be updated anyway.

Not my project, though. I'll shut up.

TingPing:
@vgturtle127 Do realize that this API is well over a decade old and those scripts largely still work.

vgturtle127:
@TingPing Yeah, my point is you shouldn't hold back from updating just because it is old, or just because it will break everything. I have been involved with projects that had that mindset and suffered greatly because of it.

I'm sure the API is fine, I am just saying 10 years of stale development may make the API less preferable in the future.

FichteFoll:
>my point is you shouldn't hold back from updating just because it is old, or just because it will break everything.

It was explained that the was considered *too minor* to warrant breaking the API, including all themes and scripts that relied on it.

vgturtle127:
@FichteFoll I know, I suppose there is no use all of us saying the same things over and over. :wink:

Poorchop:
This is very unfortunate.

PR #1668, opened 2016-04-01
[RFC] Undo tab close

etrunko:
I always find it annoying when I close a channel tab by mistake and then have to join it again manually. Especially when the channel you just closed have you have long or weird names.

This request implements the action of undo tab close, as found in most web browsers today, by pressing CTRL+Shift+T.

The first patch fixes a small issue with key event handling. I think no one else had this problem before because there are no accelerators with the shift key.

The second one implements the feature for Gtk gui and it is a ***draft***. The goal is to have a stack of recently closed channels, and not be restricted by only one channel. I would appreciate some advise of the best place to store such stack, if in a global variable or as a member of a data struct.

Comments appreciated.


TingPing:
I'm not sure the case change is the right thing to do..

src/fe-gtk/maingui.c
```diff
@@ -1091,6 +1097,13 @@ mg_tab_close (session *sess)
 	GSList *list;
 	int i;
 
+	if (is_channel(sess->server, sess->channel))
+	{
+		_last_chan.server = sess->server;
+		memcpy(_last_chan.channel, sess->channel, CHANLEN);
+		memcpy(_last_chan.key, sess->channelkey, 64);
```
TingPing: `g_strlcpy()`

etrunko:
@TingPing I tried without success to match Ctrl+Shift+T. The event will always have the code of the uppercase letter, while the stored value will always be lowercase. What do you suggest?

src/fe-gtk/maingui.c
```diff
@@ -1084,6 +1084,12 @@ mg_tab_close_cb (GtkWidget *dialog, gint arg1, session *sess)
 	}
 }
 
+static struct {
+    server *server;
+    char channel[CHANLEN];
+    char key[64];
+} _last_chan = { 0 };
```
TingPing: I think it makes more sense to use a `GQueue` and store a few previous channels rather than a single one. Also don't start variables with `_`.

etrunko: Yes, that is what I intended to do, as I wrote in the description of the pull req. I need some advice of where it should be stored.

src/fe-gtk/maingui.c
etrunko: done

TingPing:
> What do you suggest?

Will have to look into it, just doesn't sound correct, keys *are* case sensitive in my mind.

Arnavion:
(Note: Haven't actually tested your code, and don't have HC in front of me to test atm.)

By ignoring the case completely, aren't you allowing the reverse to happen? i.e. <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>T</kbd> will end up triggering shortcuts defined for <kbd>Ctrl</kbd><kbd>T</kbd> ?

I think instead of ignoring completely you want to check if the keyval is uppercase, then see if there's a corresponding shortcut for <kbd>Shift</kbd><kbd>corresponding lowercase</kbd> and trigger it, else ignore it.

TingPing:
BTW I wrote this a while ago (just realized it doesn't handle keys :P): https://github.com/TingPing/plugins/blob/master/HexChat/undoclose.py

src/fe-gtk/maingui.c
TingPing: Probably not in the maingui. I don't really know the ideal spot but for example the lastact feature in fkeys.c shoved a list in `hexchat.c`, don't know how awful that location is too.

etrunko:
@Arnavion I have tested it here, it will only work for Ctrl+Shift+T. I guess you wanted to check if this change would mess with new tab when user has Caps-lock turned on right?

etrunko:
I have updated patch 2 implementing it as a GQueue as suggested. Still missing the right place to store the pointer to that queue. It will leak some memory when you close the application with channel information stored.

etrunko:
Updated patch to to avoid adding channels to the queue while hexchat is quitting.

etrunko:
Another update, use g_new/g_free instead of calloc/free.

I am thinking about having this list global, in hexchat.c. Does this solution have a chance to go in? What about the fix for accelerators?

etrunko:
New version of patch, the only difference is that we now clear closed channels when hexchat is quitting. I tried making the queue global, but did not like the result. Take a look https://gist.github.com/etrunko/a2dafb25c0ca49175c2b57b6f7011de9

etrunko:
ping?!

etrunko:
Hi @TingPing @Arnavion, any additional comments?

abnerf:
+1

etrunko:
Hi, another rebase of the code is up. Please take a look. It has been long time since any project developers replied to this pull request. I don't see any difficulties in having these patches merged, as the code is not intrusive, and I have been using it since I started working on it, but it is getting kind of painful to keep it out of the tree. As it happens with many other different open source projects out there, this kind of attitude unresponsiveness helps keeping people away from contributing and eventually using the project, unfortunately.

Arnavion:
Since you're disallowing reopening a channel tab if its server tab has been closed (since you check `is_server()`), why not scope the queue to the server object?

Note that TingPing's script [that he linked above](https://github.com/hexchat/hexchat/pull/1668#issuecomment-204551180) does allow it. I think doing it the same way is the better approach, i.e., store just network + channel + key in the queue, and connect to the network if an existing connection is unavailable.

src/fe-gtk/maingui.c
```diff
@@ -1119,6 +1145,21 @@ mg_tab_close (session *sess)
...
+void
+mg_undo_tab_close (session *sess)
+{
+	closed_channel *chan;
+
+	if (g_queue_is_empty(&closed_channels))
+		return;
+
+	chan = g_queue_pop_head(&closed_channels);
+	if (is_server (chan->server))
```
Arnavion: It's a little suboptimal to silently ignore the undo command if this is false. The user will have no idea why nothing happened, and if they press it again expecting it didn't go through the first time, it might reopen a completely different tab.

src/fe-gtk/maingui.c
```diff
@@ -1084,13 +1084,39 @@ mg_tab_close_cb (GtkWidget *dialog, gint arg1, session *sess)
...
 mg_tab_close (session *sess)
 {
 	GtkWidget *dialog;
 	GSList *list;
 	int i;
 
+	if (hexchat_is_quitting)
+	{
+		closed_channel *chan;
+		while (!g_queue_is_empty(&closed_channels))
```
Arnavion: I don't see the point of freeing memory when the process is exiting.

Arnavion: And in any case, [g_queue_free_full](https://developer.gnome.org/glib/stable/glib-Double-ended-Queues.html#g-queue-free-full) exists for this purpose.

etrunko: About g_queue_free_full, isn't it intended to be used only for dynamically allocated queue objects? I thought it was the same as g_queue_free, so I avoided using it. I need to investigate a bit further.

etrunko:
Hello there, I had not touched this until recently when I upgraded my distro and noticed this feature did not work anymore.

@Arnavion regarding scoping of the queue, I tried to do it with this new patch, but I lose the global behavior. https://gist.github.com/etrunko/c6a38e5b84227e06e092a16e7df5fe67

For instance if I have more than one server connected, and close a channel in one server, then focus on another channel of a different server, it will not restore the closed tab. I need to focus again on that other server and hit Ctrl+Shift+Tab to get that channel back open.

My idea is to reopen the last channel closed, independently of which server is focused, so I think it makes more sense to keep it global, but if you don't like the static declaration, I could easily put it under another struct, not sure which one though. Because there is no such thing as a maingui struct, I thought it would not make sense to do it like that.

Finally as for making this code work as the plugin one, I will take a look on the python bindings for the "command" function to see what it does.

src/fe-gtk/maingui.c
etrunko: Yes, I have just checked glib source code and g_queue_free_full will indeed call g_queue_free, so it should not be used in this case.

etrunko:
Updated branch with a commit adding shortcut for tab closing. Recent versions have disabled it somehow.

TingPing:
They can already run the command `/close` I don't think it needs a new keyboard action.

etrunko:
Well, for me it is something intrinsic to any GUI application which has tabs, such as web browsers, terminal emulators and chat applications. All of them support `Ctrl+w` and `Crtl+Shift+t` shortcuts, and that is really what this is all about, shortcuts. `/close` and other commands are really for people who know how to use IRC, and the user interface should provide the ways for people who are not aware those commands. For instance, I can easily change my nickname with `/nick`, but it is also possible to do it by clicking on the button on the main window. Hexchat itself supports `Ctrl+t` for a new tab, so I think it makes perfect sense to provide shortcuts for closing and "unclosing" tabs.

I wonder what really holds this change to be merged as is, and as I had already asked before, please provide me with pointers about what should be improved/changed to get this code in, so I can get rid of maintaining a separate branch and building manually.

PR #1672, opened 2016-04-04
Some more refactoring. Fix unload hook in the interpreter.

mniip:


plugins/lua/lua.c
```diff
@@ -1262,36 +1262,31 @@ static void prepare_state(lua_State *L, script_info *info)
...
 	if(lua_pcall(L, 0, 0, base))
 	{
 		char const *error = lua_tostring(L, -1);
-		lua_pop(L, 2);
 		hexchat_printf(ph, "Lua error in unload hook: %s", error ? error : "(non-string error)");
 	}
+	lua_settop(L, base);
+}
+
+static void run_unload_hooks(script_info *info, void *unused)
```
TingPing: Reason for having `unused`?

Arnavion: `/* The unused parameter is present because this function is used as a GFunc */`

PR #1681, opened 2016-04-10
Support for older libnotify versions

mtelka:


TingPing:
Curious what distribution has libnotify 0.6.0 (2011) but also has glib 2.34.0+ (2012)?

mtelka:
[OpenIndiana Hipster](http://www.openindiana.org/overview/the-hipster-branch/), currently with glib 2.43.4 and libnotify 0.4.5.

TingPing:
Those are quite the odd package versions, a 7yo libnotify and a year old development release of glib... but fair enough

TingPing:
Merged as 54c6418d619e1ec96166f9fabb426f011068de99

PR #1704, opened 2016-05-08
Improve timer behavior for more effective power management

bgamari:
Currently Hexchat is a remarkable power hog. Not only does it make no attempt to consolidate wake-ups (to maximize processor sleep state residency) but it also has a number timers that are entirely superfluous. For this reason Hexchat is consistently one of the biggest wake-up sources on my laptop. This can be observed using either `powertop` or `/proc/timer_stats` directly.

This branch is an attempt to fix this. It does a few things,

 * break up the `hexchat_misc_checks` callback to separate unrelated timers
 * disable the lag-o-meter timer entirely when this feature is not in use
 * use `g_timeout_add_seconds` wherever possible to allow the operating system to choose wake-up times more appropriately
 * disable the DCC timeout timer when there are no DCC connections active
 * a bit of miscellaneous refactoring in DCC-land

One remaining violator that I don't really know how to handle is the `python` plugin, which registers a 300ms timer to allow Python's threading mechanism to do its thing. It's not really clear to me what to do about this and I don't use this plugin myself.

After unloading `python` these changes place `hexchat` well down the list of wake-up sources, where it belongs.

### DIFF
--- a/src/common/cfgfiles.h
+++ b/src/common/cfgfiles.h
@@ -71,6 +71,11 @@ struct prefs
 	unsigned short offset;
 	unsigned short len;
 	unsigned short type;
+	/*
+	 * an optional function which will be called after the preference value has
+	 * been updated.
+	 */
+	void (*after_update)(void);
 };
 
 #define TYPE_STR 0
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -354,13 +354,6 @@ doover:
 	return 1;
 }
 
-static int
-hexchat_check_dcc (void)	/* this gets called every 1 second */
-{
-	dcc_check_timeouts ();
-	return 1;
-}
-
 /* these are only run if the lagometer is enabled */
 static int
 hexchat_lag_check (void)   /* this gets called every 30 seconds */
@@ -406,7 +399,6 @@ irc_init (session *sess)
 					     notify_checklist, NULL);
 
 	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, NULL);
-	fe_timeout_add (1000, hexchat_check_dcc, NULL);
 	if (prefs.hex_gui_lagometer)
 	{
 		fe_timeout_add (500, hexchat_lag_check_update, NULL);
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -369,6 +369,37 @@ hexchat_lag_check_update (void)   /* this gets called every 0.5 seconds */
 	return 1;
 }
 
+/* call whenever timeout intervals change */
+void
+hexchat_reinit_timers (void)
+{
+	static int lag_check_update_tag = 0;
+	static int lag_check_tag = 0;
+	static int away_tag = 0;
+
+	if (prefs.hex_notify_timeout)
+	{
+		notify_tag = fe_timeout_add_seconds (prefs.hex_notify_timeout,
+						     notify_checklist, NULL);
+	} else
+	{
+		fe_timeout_remove (notify_tag);
+	}
+
+	fe_timeout_remove (away_tag);
+	away_tag = fe_timeout_add_seconds (prefs.hex_away_timeout, away_check, NULL);
+
+	if (prefs.hex_gui_lagometer)
+	{
+		lag_check_update_tag = fe_timeout_add (500, hexchat_lag_check_update, NULL);
+		lag_check_tag = fe_timeout_add_seconds (30, hexchat_lag_check, NULL);
+	} else
+	{
+		fe_timeout_remove (lag_check_update_tag);
+		fe_timeout_remove (lag_check_tag);
+	}
+}
+
 /* executed when the first irc window opens */
 
 static void
@@ -394,16 +425,7 @@ irc_init (session *sess)
 	plugin_add (sess, NULL, NULL, dbus_plugin_init, NULL, NULL, FALSE);
 #endif
 
-	if (prefs.hex_notify_timeout)
-		notify_tag = fe_timeout_add_seconds (prefs.hex_notify_timeout,
-						     notify_checklist, NULL);
-
-	fe_timeout_add_seconds (prefs.hex_away_timeout, away_check, NULL);
-	if (prefs.hex_gui_lagometer)
-	{
-		fe_timeout_add (500, hexchat_lag_check_update, NULL);
-		fe_timeout_add_seconds (30, hexchat_lag_check, NULL);
-	}
+	hexchat_reinit_timers ();
 
 	if (arg_url != NULL)
 	{
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -395,14 +395,14 @@ irc_init (session *sess)
 #endif
 
 	if (prefs.hex_notify_timeout)
-		notify_tag = fe_timeout_add (prefs.hex_notify_timeout * 1000,
-					     notify_checklist, NULL);
+		notify_tag = fe_timeout_add_seconds (prefs.hex_notify_timeout,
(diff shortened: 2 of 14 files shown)


TingPing:
At first glance:

- The dcc_type commit seems entirely unrelated to this set of changes
- Put braces on their own line, [Allman style](https://en.wikipedia.org/wiki/Indent_style#Allman_style)
- Use NULL rather than 0 for pointer args

bgamari:
Are the project style guidelines documented anywhere? I've been unable to find them.

bgamari:
I've split out the unrelated commit into #1705 and addressed your comments in the remaining commits.

src/common/hexchat.c
```diff
@@ -403,10 +403,15 @@ irc_init (session *sess)
...
 	if (prefs.hex_notify_timeout)
 		notify_tag = fe_timeout_add (prefs.hex_notify_timeout * 1000,
-											  notify_checklist, 0);
+					     notify_checklist, NULL);
 
-	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, 0);
-	fe_timeout_add (500, hexchat_misc_checks, 0);
+	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, NULL);
+	fe_timeout_add (1000, hexchat_check_dcc, NULL);
+	if (prefs.hex_gui_lagometer)
```
TingPing: This means you have to restart the client to enable that feature, less than ideal.

src/common/hexchat.c
```diff
@@ -403,10 +403,15 @@ irc_init (session *sess)
...
+	{
+		fe_timeout_add (500, hexchat_lag_check_update, NULL);
+		fe_timeout_add (30000, hexchat_lag_check, NULL);
```
TingPing: This is used to detect timeouts not for the lagmeter. This can also be disabled (and is by default) but the same comment about requiring restarting applies.

src/common/hexchat.c
bgamari: TingPing <notifications@github.com> writes:

>>  
>> -	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, 0);
>> -	fe_timeout_add (500, hexchat_misc_checks, 0);
>> +	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, NULL);
>> +	fe_timeout_add (1000, hexchat_check_dcc, NULL);
>> +	if (prefs.hex_gui_lagometer)
>
> This means you have to restart the client to enable that feature, less than ideal.
>
Indeed; I've pushed one possible fix for this.

src/common/hexchat.c
bgamari: TingPing <notifications@github.com> writes:

>>  
>> -	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, 0);
>> -	fe_timeout_add (500, hexchat_misc_checks, 0);
>> +	fe_timeout_add (prefs.hex_away_timeout * 1000, away_check, NULL);
>> +	fe_timeout_add (1000, hexchat_check_dcc, NULL);
>> +	if (prefs.hex_gui_lagometer)
>> +	{
>> +		fe_timeout_add (500, hexchat_lag_check_update, NULL);
>> +		fe_timeout_add (30000, hexchat_lag_check, NULL);
>
> This is used to detect timeouts not for the lagmeter. This can also be
> disabled (and is by default) but the same comment about requiring
> restarting applies.
>
Also fixed in latest branch.

Cheers,

- Ben

src/common/hexchat.c
```diff
@@ -369,6 +369,37 @@ hexchat_lag_check_update (void)   /* this gets called every 0.5 seconds */
...
+	{
+		fe_timeout_remove (notify_tag);
+	}
+
+	fe_timeout_remove (away_tag);
+	away_tag = fe_timeout_add_seconds (prefs.hex_away_timeout, away_check, NULL);
+
+	if (prefs.hex_gui_lagometer)
+	{
+		lag_check_update_tag = fe_timeout_add (500, hexchat_lag_check_update, NULL);
```
TingPing: This would result in many timers being created, if there is already a tag it should do nothing, also make sure to set tags to 0 when disabled.

src/common/hexchat.c
```diff
@@ -369,6 +369,37 @@ hexchat_lag_check_update (void)   /* this gets called every 0.5 seconds */
...
+		lag_check_tag = fe_timeout_add_seconds (30, hexchat_lag_check, NULL);
```
TingPing: As said before this is for network timeouts not just lagmeter.

src/fe-gtk/setup.c
```diff
@@ -2132,6 +2132,7 @@ setup_apply_real (int new_pix, int do_ulist, int do_layout, int do_identd)
 
 	mg_apply_setup ();
 	tray_apply_setup ();
+	hexchat_reinit_timers ();
```
TingPing: Note that `hex_gui_lagometer` can be toggled from the menu and that `hex_net_ping_timeout` is not exposed anywhere so that will no longer work.

src/common/hexchat.c
```diff
@@ -369,6 +369,37 @@ hexchat_lag_check_update (void)   /* this gets called every 0.5 seconds */
...
+	{
+		notify_tag = fe_timeout_add_seconds (prefs.hex_notify_timeout,
+						     notify_checklist, NULL);
+	} else
+	{
+		fe_timeout_remove (notify_tag);
+	}
+
+	fe_timeout_remove (away_tag);
+	away_tag = fe_timeout_add_seconds (prefs.hex_away_timeout, away_check, NULL);
```
TingPing: This one can also be conditional behind `hex_away_track` I believe.

src/fe-gtk/setup.c
bgamari: Not entirely sure what you mean regarding `hex_net_ping_timeout`.

TingPing: The preference variable used in `hexchat_check_lag()`, it is only exposed via `/set` which means it now cannot be changed without restarting. It really isn't your fault but I think ideally this would just be exposed in the preferences window.

bgamari:
@TingPing I've added a couple patches adding a mechanism with which preferences can handle setting changes and using this mechanism to reinitialize timers when `net_ping_timeout` is changed.

bgamari:
Pinging @TingPing.

TingPing:
Looks good overall. Not sure how I feel about the after_update work because that leads to a larger refactoring across all of hexchat... :P

src/common/cfgfiles.h
```diff
@@ -71,6 +71,11 @@ struct prefs
 	unsigned short offset;
 	unsigned short len;
 	unsigned short type;
+  /*
+   * an optional function which will be called after the preference value has
+   * been updated.
+   */
+  void (*after_update)(void);
```
TingPing: Tabs for whitespace.

bgamari: TingPing <notifications@github.com> writes:

>> @@ -71,6 +71,11 @@ struct prefs
>>  	unsigned short offset;
>>  	unsigned short len;
>>  	unsigned short type;
>> +  /*
>> +   * an optional function which will be called after the preference value has
>> +   * been updated.
>> +   */
>> +  void (*after_update)(void);
>
> Tabs for whitespace.
>
Fixed.

bgamari:
Do you need anything else on this?

bgamari:
So is this ready for merge?

TingPing:
I'll try to soon.

TingPing:
Merged with a few modifications, thanks!

Issue #500, opened 2013-04-04
Improve ignore feature


haarp:
Hello,

first off, thank you for maintaining and improving Xchat!

I have a small feature request for something that is, in my opinion, sorely lacking. The ignore feature allows you to ignore channel messages of certain people, but you still see other interaction. That includes joins, parts, quits, nick changes, mode changes, etc. would it be possible to add a checkbox to block these from view as well?

It would be much appreciated. Thanks!

duckmaestro:
I'd also like to see this. We're using a service which likes to constantly issue "* | John gives voice to John" and "* | John removes voice from John" messages based on activity/idle state or other sensors. I'd like to be able to ignore these messages on certain channels.

Arnavion:
@duckmaestro What you are asking for is closer to #349 (which contains a workaround), whereas this one is for ignoring all messages related to *an already ignored user.*

PR #1712, opened 2016-05-13
Added ability to ignore joins, parts and quits of an ignored user.

Arnavion:
Mentioned in #500 

---

Note that only the text events emit is skipped, to keep HC's state of the channel sane. For example, the ignored user will still be added to the channel list when they join and be available for tab-complete, etc. which would not be the case if the JOIN had been ignored in inbound.c. Otherwise you could end up in strange states, such as if the ignored user was present in the channel list before being ignored, and now would not be removed even if they quit.

### DIFF
--- a/src/common/ignore.c
+++ b/src/common/ignore.c
@@ -151,6 +151,10 @@ ignore_showlist (session *sess)
 			strcat (tbuf, _("YES  "));
 		else
 			strcat (tbuf, _("NO   "));
+		if (ig->type & IG_JOINS_PARTS)
+			strcat (tbuf, _("YES         "));
+		else
+			strcat (tbuf, _("NO          "));
 		strcat (tbuf, "\n");
 		PrintText (sess, tbuf);
 		/*EMIT_SIGNAL (XP_TE_IGNORELIST, sess, ig->mask, 0, 0, 0, 0); */
--- a/src/common/ignore.c
+++ b/src/common/ignore.c
@@ -77,7 +77,7 @@ ignore_exists (char *mask)
  */
 
 int
-ignore_add (char *mask, int type, gboolean overwrite)
+ignore_add (char *mask, unsigned int type, gboolean overwrite)
 {
 	struct ignore *ig = NULL;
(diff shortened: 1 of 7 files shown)


src/common/ignore.c
```diff
@@ -77,7 +77,7 @@ ignore_exists (char *mask)
  */
 
 int
-ignore_add (char *mask, int type, gboolean overwrite)
+ignore_add (char *mask, unsigned int type, gboolean overwrite)
```
TingPing: Why not `guint`?

src/common/inbound.c
```diff
@@ -796,18 +797,19 @@ inbound_kick (server *serv, char *chan, char *user, char *kicker, char *reason,
...
 	if (sess)
 	{
-		if (*reason)
-			EMIT_SIGNAL_TIMESTAMP (XP_TE_PARTREASON, sess, user, ip, chan, reason,
-										  0, tags_data->timestamp);
-		else
-			EMIT_SIGNAL_TIMESTAMP (XP_TE_PART, sess, user, ip, chan, NULL, 0,
-										  tags_data->timestamp);
+		if (!ignore_check(host, IG_JOINS_PARTS))
+			if (*reason)
```
TingPing: Probably should put braces around this.

src/common/ignore.c
```diff
@@ -151,6 +151,10 @@ ignore_showlist (session *sess)
 			strcat (tbuf, _("YES  "));
 		else
 			strcat (tbuf, _("NO   "));
+		if (ig->type & IG_JOINS_PARTS)
+			strcat (tbuf, _("YES         "));
```
TingPing: Inconsistent whitespace compared to the others?

src/common/ignore.c
Arnavion: Sure, why not.

src/common/ignore.c
Arnavion: It needs to match the header (`XP_TE_IGNOREHEADER`).

PR #1771, opened 2016-07-13
lua: Add automatic return and = handling in console.

mniip:


plugins/lua/lua.c
```diff
@@ -1481,15 +1481,30 @@ static void inject_string(script_info *info, char const *line)
 {
 	lua_State *L = info->state;
 	int base, top;
+	char *program, *ret_program;
+
+	if(line[0] == '=')
+		program = g_strconcat("return ", line + 1, NULL);
+	else
+		program = g_strdup(line);
+	ret_program = g_strconcat("return ", program, NULL);
```
Arnavion: Do you really want to do `return return` or do you just want to do

```
1. if line[0] == '=' then line = line + 1
2. program = "return " + line
```

Arnavion: Also how does this interact with multi-line lua statements. Do you need to wrap `line` in something like an IIFE and return that instead?

mniip:
`return return` is not an issue because that is obviously a syntactic error and the first loadbuffer will fail and the other one will be used.

Arnavion:
I'm trying to say that rather than having the possibilty of constructing a `return return` that's guaranteed to not be what you want, you could just strip the leading `=`

mniip:
Would that make much difference? Seems like it would only complicate the logic by a bunch.

Arnavion:
Umm...

```c
if(line[0] == '=')
	program = g_strconcat("return ", line + 1, NULL);
else
	program = g_strdup(line);
ret_program = g_strconcat("return ", program, NULL);
```

becomes

```c
if(line[0] == '=')
	line = line + 1;
ret_program = g_strconcat("return ", line, NULL);
```

mniip:
Hmm I suppose.

Arnavion:
This still breaks multiple lines separated by `;`, right?

mniip:
Not at all. `return x(); y()` is a syntactic error.

Issue #1774, opened 2016-07-22
Non-ident prefix "-" prevents hilighting based on recipient nick.


kfogel:
This is with latest HexChat (as of commit ba8724778):

If speaker Q hasn't identified to the server, then sometimes everything Q says appears to listeners as prefixed with a leading '-' (whether this happens or not depends on various settings).  But suppose that Q is addressing HexChat user N, and thus Q begins each line with N's nick (as is common etiquette in some chat rooms).   Then when Q says something like:

    N: hi there

...what N's HexChat client actually shows is:

    -N: hi there

The bug is that the leading "-" in "-N" then causes Q's speaker nick to *not* be highlighted (for N), even if N's preferences say that such highlighting should happen for any line in which someone says the word "N"... because HexChat doesn't see the word "N", but instead sees the word "-N".

Proposed solution:

In `is_hilight()`, just strip off a single leading "-" from the text before trying to match it against masks for hilighting purposes.  It would be extremely rare that both of these things are true: the first character of the first word of a line is a hyphen *and* it be important to the recipient that that hyphen-prepended word not match for hilighting while its non-hyphened counterpart should match.  

I have a patch for that behavior and will submit the pull request in a moment.

(A somewhat narrower solution would be to detect if any of the masks care about a leading hyphen, and behave appropriately, but I doubt that's worth the extra complexity.  After all, https://tools.ietf.org/html/rfc2812#section-2.3.1 specifies that '-' isn't allowed as the first character of a nickname, and 99% of the time, if there's going to be a match for that first word, it's going to be because it's the recipient's nickname.  In any case, most likely that someone who wants "N" matched probably also wants "-N" matched whether or not N is their nick, since lone leading hyphens are rare and usually do not represent a distinct word as compared to the same word without the hyphen.)

kfogel:
See PR #1775.

TingPing:
Sounds like you have a script that is requesting `identify-msg` behind HexChat's back.

TingPing:
This was confirmed on IRC:

> sasl-lite-xchat.pl 1.0
cap_sasl_xchat.pl 1.0300
cap_sasl_xchat.py 0.5

These request caps behind hexchats back and only cause problems. HexChat has SASL in the network list (ctrl+s > edit) so these are no longer needed.

PR #1775, opened 2016-07-22
Fix hilight in case of non-ident prefix '-' #1774

kfogel:


TingPing:
https://github.com/hexchat/hexchat/issues/1774#issuecomment-234590654

PR #1779, opened 2016-07-26
count code contributors per organization

dachary:
The FLOSS Contributor organization groups all contributors that are not
affiliated to any known organization.

Signed-off-by: Loic Dachary <loic@dachary.org>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Patrick Griffis <tingping@tingping.se> <TingPing@users.noreply.github.com>
--- a/.organizationmap
+++ b/.organizationmap
@@ -1,0 +1,29 @@
+#
+# Display the number of contributors active in the past six months, per organization
+#
+# git log --no-merges --pretty='%aN <%aE>' --since '6 months ago' | sort | uniq | git -c mailmap.file=.organizationmap check-mailmap --stdin | sort | uniq -c | sort -rn | nl
+#     1	     16 FLOSS Contributor <contact@floss.cc>
+#
+# Display the number of commits in the past six months, per organization
+#
+# git log --no-merges --pretty='%aN <%aE>' --since '6 months ago' | git -c mailmap.file=.organizationmap check-mailmap --stdin | sort | uniq -c | sort -rn | nl
+#     1	    113 FLOSS Contributor <contact@floss.cc>
+#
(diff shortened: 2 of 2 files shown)


TingPing:
What is this and why do we need it?

dachary:
@TingPing it's a census of the contributors to hexchat and the organizations they belong to. It shows there only are volunteers and noone is sponsored by a company. Other projects such as Ceph https://github.com/ceph/ceph/blob/master/.organizationmap are mostly sponsored by companies.

TingPing:
I don't think this is particularly useful information to store in its own file. The git log contains all of this info and doesn't require somebody maintaining a file.

dachary:
> The git log contains all of this info
How can you figure that out just be looking at the git log ? I certainly could not reach the conclusion proposed in this pull request without looking at the web site and actually understanding where hexchat comes from. Is there maybe a reason why you would not want this information to be made public ?

TingPing:
I admit it requires some effort and knowledge to use git but honestly I don't find this information particularly relevant. This is just a small project with a handful of active developers unlike a large commercial project like Ceph.

dachary:
I understand your point of view, thanks for explaining :-) Why not just welcome this contribution, even if from your perspective it seems marginally useful ? I would expect a Free Software project to appreciate the effort of contributors and embrace diverse view points, as long as it does not jeopardize the stability of the project or otherwise negatively impact it.

TingPing:
Well somebody has to keep the information updated otherwise it becomes incorrect information which is worse than none.

dachary:
I should have clarified the maintenance aspect in the **.organizationmap** comment, this is a very good point. 

A census is meant to be true at a given point in time (July 26th, 2016 in this case) and verifiable (which is why I included the command lines). I overlooked the fact that someone could misinterpret these files as a way to magically create an accurate census of the developers affiliations from the git logs at any point in time although the git logs do no contain the required information.

You will probably think that a single census is even less useful :-) The census result is however included in http://wikidata.org and the census method is documented so that it can be crowdsourced in the long run. I do my best to increase the probability of having another census of the hexchat in the future, as part of the routine maintenance the wikidata.

Should you agree to accept this census, I'll update the comment to explain what to expect in terms of maintenance.

PR #1790, opened 2016-08-11
Added MHOP command

d-browne:
Noticed that there is no MHOP command. Given that there is a MOP, MDEOP and MDEHOP it seems like the useful MHOP command is missing. I have added it in.

### DIFF
--- a/src/common/outbound.c
+++ b/src/common/outbound.c
@@ -1338,6 +1338,33 @@ cmd_menu (struct session *sess, char *tbuf, char *word[], char *word_eol[])
 }
 
 static int
+mhop_cb (struct User *user, multidata *data)
+{
+	if (!user->hop)
+	{
+		data->nicks[data->i] = user->nick;
+		data->i++;
+	}
+	return TRUE;
+}
+
+static int
+cmd_mhop (struct session *sess, char *tbuf, char *word[], char *word_eol[])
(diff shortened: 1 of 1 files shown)


TingPing:
Half op is not universally supported and AFAIK none of the other built in commands apply to half-op so I'm not sure we should start now. A script could ofc trivially do this.

d-browne:
Thank you TingPing. There are actually three built in half-op commands already. HOP, DEHOP and MDEHOP. The only one that is missing is MHOP which is why I thought it appropriate to add. Especially given that MDEHOP is already there.

TingPing:
Ah so there is. Another question is when is mass oping even wanted?

d-browne:
The three situations I've needed mass oping and hoping have usually been related to: the creation of new channels and quickly oping the founders/channel bots; initializing channel wide ops for all-op channels; and channel festivities/games.

Funnily enough two people in one of my channels noticed the mhop feature missing from hexchat at the same time, shortly after trying to restore the channel wide half ops after a mdehop. That inspired me to head over here :]

TingPing:
> (needs chanop)

Can't a halfop halfop other users?

kernelpanic3:
No, they cannot (at least that's the way it is on all the networks I've been on).

Issue #739, opened 2013-09-12
Hexchat 2.9.6 broke autojoin channels.


jordanmkasla2009:
Upon starting hexchat 2.9.6 after upgrading i noticed that it was only joining one channel on my autojoin list. i looked it up to find that the first channel was there but the rest of the channels were inserted into the key of the first channel.

PR #1794, opened 2016-08-16
Add support for migration from xchat configuration

scarabeusiv:
If user starts hexchat for the first time and yet he still has some
xchat2 configuration around we should reuse it thus saving the user
some time.

We in SUSE/openSUSE switched the default from xchat2 to hexchat and thus want the users to be able to use their irc right after the distribution update.

### DIFF
--- a/src/common/cfgfiles.c
+++ b/src/common/cfgfiles.c
@@ -950,6 +950,38 @@ make_dcc_dirs (void)
 	return 0;
 }
 
+void migrate_from_xchat (void)
+{
+	char *servlist_conf, *logs_xchat;
+	char *xchat_dir = g_build_filename (g_get_home_dir (), ".xchat2", NULL);
+	if (!g_file_test (xchat_dir, G_FILE_TEST_EXISTS))
+	{
+		g_free (xchat_dir);
+		return;
+	}
+	servlist_conf = g_build_filename (xchat_dir, "servlist_.conf", NULL);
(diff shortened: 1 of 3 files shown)


Arnavion:
https://github.com/hexchat/hexchat/issues/739#issuecomment-24291893

and probably all sorts of incompat in the prefs themselves.

TingPing:
- servlist.conf: needs manual migrating and is worse than the new defaults (SASL/SSL on many networks and removing dead ones, adding new ones)
- pevents.conf are not necessarily a good idea to bring over, there are many new events so old themes are a mix of new and old.
- addons: A lot of them are not needed any longer and could cause conflicts such as SASL scripts.
- hexchat.conf: Similiar to servlist there are a lot of new saner defaults that we want users to have but also tons of settings have been added, renamed, removed.

TingPing:
Anyway, with that said I think there are only two good options for migration.

- Don't migrate at all, it really isn't that hard to do a one time migrating and it is impossible to have a complete full migration so why half-ass it.

- Have a slightly more interactive migration process. Present a prompt showing users that you can migrate your addons and maybe a few specific configs but inform them that not everything can be migrated. That way the important bits are done and users didn't have to read docs about it but it didn't create a mess mass migrating things.

scarabeusiv:
We luckily abandoned the migration in openSUSE :)

PR #1817, opened 2016-09-18
configure: fix --disable-python

raimue:
Due to missing quotes on one side of this shell comparison the --disable-python argument is not working as expected. python2 was always included in the build.


raimue:
Sorry, I was completely wrong here. I was fooled by the fact that `autogen.sh` already runs configure without parameters unless `NOCONFIGURE` is in the environment. This actually works correctly as is.

PR #1839, opened 2016-10-11
Remove dead assignment

f2404:


TingPing:
Thanks for all these cleanups, but for future reference for things like plugins the commit message is generally prefixed with the plugin it is for. so `perl: Remove dead assignment`.

f2404:
Thanks for merging my patches!
I've rebased this one to change the message.

PR #1841, opened 2016-10-12
mpcinfo: Do not modify string literals

f2404:
Modification of string literals causes undefined behavior: http://stackoverflow.com/questions/8302290/why-does-char-cause-undefined-behaviour-while-char-doesnt

### DIFF
--- a/plugins/mpcinfo/oggInfo.c
+++ b/plugins/mpcinfo/oggInfo.c
@@ -48,7 +48,7 @@ struct tagInfo getOggHeader(char *file){
 	char *sub;
 	char *name;
 	char *val;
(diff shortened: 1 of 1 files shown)


TingPing:
btw the entire mpcinfo plugin is a garbage codebase, the only reason it hasn't been deleted is because I'm lazy.

f2404:
Yeah, the code is really awful.

TingPing:
Do you use the plugin personally or just saw tons of issues?

f2404:
I'm not using it (and as far as I can tell it's not built for linux). But yes, I saw the issues and the code style.

PR #1876, opened 2016-11-19
RFC: Lurk mode

andreamerello:
Hello,
I'm submitting this PR early just to probe for opinions about it, even if I don't consider this finished yet.

Often I end up sending messages by mistake, for example because I think another window have focus, or because I write in the wrong channel.. Since I usually lurk on several channels, and I rarely interact, I wanted an option to disable text input on certain channels on my choose, so that it become impossible to send messages by mistake (unless you disable that option, of course :) )

This PR introduces the "lurk mode" in channel settings, so that the input box became not editable when the option is active.

I would like to make this a bit better, for example making the input box grey when in lurk mode, and/or make the channel name grey for channels in lurk mode.. Also, I haven't taken care of saving this settings in the config files yet.. But as I said I'm looking for early comments about this :)

### DIFF
--- a/src/common/hexchat.h
+++ b/src/common/hexchat.h
@@ -176,6 +176,7 @@ struct hexchatprefs
 	unsigned int hex_irc_hide_version;
 	unsigned int hex_irc_invisible;
 	unsigned int hex_irc_logging;
+	unsigned int hex_irc_lurk;
 	unsigned int hex_irc_raw_modes;
 	unsigned int hex_irc_servernotice;
 	unsigned int hex_irc_skip_motd;
@@ -372,6 +373,7 @@ typedef struct session
 	guint8 text_logging;
 	guint8 text_scrollback;
 	guint8 text_strip;
+	guint8 text_lurk;
 
 	struct server *server;
 	tree *usertree;					/* alphabetical tree */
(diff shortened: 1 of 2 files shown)


TingPing:
I personally don't think this is really a core feature a lot of users would want so I played with the idea of making this a plugin:

```lua
local lgi = require('lgi')
local Gtk = lgi.require('Gtk', '2.0')

hexchat.register('LurkMode', '1', 'Disables input for a channel')

local function get_widget_at_index(widget, index)
	for _, i in ipairs(index) do
		widget = widget:get_children()[i]
	end
	return widget
end

local function set_lurk_mode(enabled)
	local window = Gtk.Window(hexchat.get_info('gtkwin_ptr'))

	-- This is ugly but this is the layout of the UI:
	local entry = get_widget_at_index(window:get_child(), { 2, 1, 2, 1, 1, 4, 2 })
	entry:set_editable(enabled == false)
--	entry:set_sensitive(enabled == false)
end

hexchat.hook_unload(function ()
	set_lurk_mode(false)
end)

set_lurk_mode(true)
``

mB-PiBox:
I dig the concept/idea, but I personally would not use it just because I, and most people I know, use monitor buffers/windows via script. Sending anything to those buffers goes no where, unless you specify a channel name at beginning of msg. My personal script runs at a ZNC level, but I think theres one or two floating around for Hexchat.

andreamerello:
@TingPing  Thank you, I will give a try to your script soon. I agree with you that a plugin would be probably better :)

@mB-PiBox Sorry I'm not so familiar with this stuff. May you explain me what do you mean with "buffers/windows via script" and what's "ZNC level" ? Thanks!

PR #1878, opened 2016-11-22
fkeys: allow set buffer to empty string

invisiblek:
* Useful for using a keyboard shortcut to clear out the buffer


TingPing:
This can already be done: Action = *Run Command*, Data1 = */settext*

invisiblek:
Ah, well I suppose that solves it for me then. Thanks =)

PR #1886, opened 2016-11-30
Bump version to 2.12.4-alpha so that we can test 3118df20422cb73444ed251197b2724252de8766

paladox:
3118df20422cb73444ed251197b2724252de8766


TingPing:
That isn't how it works.

PR #1889, opened 2016-12-01
properly clean the autogenerated org.hexchat.service.service

mapreri:


mapreri:
Btw, why is it '.service.service' ?!

Arnavion:
>Btw, why is it '.service.service' ?!

Because that's what xchat had when its code was imported.

Issue #1898, opened 2016-12-12
building with LibreSSL fails due to OpenSSL 1.1.0 related changes


woohooyeah:
Hi,

Up to version 2.12.3 I've used a PKG_CONFIG_PATH variable, in order to point to LibreSSL's pkgconfig directory and build Hexchat. This has worked for as long as I can remember, until version 2.12.4 came out.

Build currently fails here:
```
CC       libhexchatcommon_a-ssl.o
ssl.c: In function ‘_SSL_get_cert_info’:
ssl.c:182:13: error: implicit declaration of function ‘X509_get_signature_nid’ [-Werror=implicit-function-declaration]
  sign_alg = X509_get_signature_nid (peer_cert);
             ^~~~~~~~~~~~~~~~~~~~~~
ssl.c: In function ‘_SSL_socket’:
ssl.c:312:11: error: implicit declaration of function ‘SSL_CTX_get_ssl_method’ [-Werror=implicit-function-declaration]
  method = SSL_CTX_get_ssl_method (ctx);
           ^~~~~~~~~~~~~~~~~~~~~~
ssl.c:312:9: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
  method = SSL_CTX_get_ssl_method (ctx);
         ^
```

After applying the following patch, building succeeds again:
```diff
--- a/plugins/fishlim/dh1080.c	2016-10-23 09:40:01.000000000 +0200
+++ b/plugins/fishlim/dh1080.c	2016-12-12 23:26:54.533943912 +0100
@@ -74,7 +74,7 @@
 
 		BN_set_word (g, 2);
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || LIBRESSL_VERSION_NUMBER
 		g_dh->p = p;
 		g_dh->g = g;
 #else
@@ -162,7 +162,7 @@
 		return 0;
 	}
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || LIBRESSL_VERSION_NUMBER
 	dh_pub_key = dh->pub_key;
 	dh_priv_key = dh->priv_key;
 #else
@@ -213,7 +213,7 @@
 
 	  	priv_key_data = dh1080_decode_b64 (priv_key, &priv_key_len);
[...]

PR #1899, opened 2016-12-13
Use AC_CHECK_FUNCS to find functions not in LibreSSL

n0la:
LibreSSL might not have all functions of OpenSSL 1.1.0 so use AC_CHECK_FUNCS to find them first before using them.

This should fix LibreSSL builds as described in #1898.


TingPing:
Doesn't include the checks for the fish plugin.

n0la:
This builds fine on OpenBSD, granted I didn't test anything else. LibreSSL misses the functions but not the members of the DH structure, so I not all parts of the original patch are actually needed. Care to elaborate what you mean?

TingPing:
I was just going off of #1898 which changed the plugin. It is possible they did that without testing it.

PR #1917, opened 2017-01-02
Configure display of plugin load/unload messages.

AndersBillLinden:
This change adds a checkbox "Display plugin/extension load/unload messages" under Settings->preferences->Chatting->Developer (also a new section).

If the checkbox is unchecked, those startup messages (like "Perl interface loaded") will not show up.

### DIFF
--- a/plugins/checksum/checksum.c
+++ b/plugins/checksum/checksum.c
@@ -38,6 +38,7 @@ static hexchat_plugin *ph;									/* plugin handle */
 static char name[] = "Checksum";
 static char desc[] = "Calculate checksum for DCC file transfers";
 static char version[] = "3.1";
+static int show_plugin_messages = 0;
 
 static void
 set_limit (char *size)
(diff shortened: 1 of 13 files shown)


AndersBillLinden:
Some changes are missing. I need to update the branch before you take a look. Sorry, forkhub seems to not be able to close pull requests!

TingPing:
Not entirely sure how I feel about this. To be meaningful every addon ever made would have to check this value which obviously cannot happen.

AndersBillLinden:
That is correct, maybe hexchat_printf should be filtered instead?

AndersBillLinden:
Another obvious way is to not let the loading and unloading messages be a responsibility of respective plugin.

TingPing:
I think filtering prints is the wrong way to go, magical non-obvious behavior.

I'm not necessarily against standardizing a load/unload message but to be effective it still means every pre-existing script must remove their messages.

Arnavion:
I don't see the point.

AndersBillLinden:
The point is that I wanted to categorize the output as something that only developers are interested in.

TingPing:
Honestly I think it is the opposite, it is only useful for users. If I am debugging an issue figuring out when a script loads isn't the hard part. For users though when they type `/load` they expect feedback or they are confused.

AndersBillLinden:
Yeah, if they type /load, they should get the output, but I get it when I start hexchat.

AndersBillLinden:
Also, if they tick that box, they will get the output when they use /load.

PR #1937, opened 2017-01-22
Do not append to user variables CFLAGS and LDFLAGS

nvinson:
Per GNU automake documentation recommendations, user variables should
not be appended to.  This allows the user final say over the
configuration without removing compiler flags needed to properly build
hexchat.

To properly follow this recommendation, hexchat_def_CFLAGS and
hexchat_def_LDFLAGS are defined and populated with the flags that were
previously appended CFLAGS and LDFLAGS.  AM_CFLAGS and AM_LDFLAGS along
with the various "mumble_CFLAGS" and "mumble_LDFLAGS" are updated to use
the hexchat_def_CFLAGS and hexchat_def_LDFLAGS.  The end result is that
any user building hexchat can now set CFLAGS and LDFLAGS to control the
build.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,6 @@
 ## Process this file with automake to produce Makefile.in
+AM_CFLAGS = @hexchat_def_CFLAGS@
+AM_LDFLAGS = @hexchat_def_LDFLAGS@
 
 ACLOCAL_AMFLAGS = -I m4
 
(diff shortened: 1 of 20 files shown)


nvinson:
[Gentoo bug #606522](https://bugs.gentoo.org/show_bug.cgi?id=606522) was the motivation for this pull request.

PR #1940, opened 2017-01-25
Add a new method to the plugin API

culb:
Add a new method to the plugin API. This will send a specified string directly to the server.


TingPing:
So again, what exactly is the use case that `/quote` doesn't handle?

PR #1941, opened 2017-01-29
Respect browser settings regarding new window creation for the default url handler

dessant:
When `-new-window` is used sometimes Firefox won't start on Windows, or the url is not opened in an already running browser instance. Removing the option fixes this bug, and also ensures that browser settings for opening new urls in a new tab or window are respected.

https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#-url_URL


TingPing:
That is sort of on purpose, the message is even `"Open Link in a new Firefox Window"`.

Also realize that this entry is little more than an example of how you might use a url handler and can simply be modified in *Settings > URL Handlers*. It also has no effect at all on normally opening urls.

PR #1942, opened 2017-01-30
Add plugin_pref_del to the Perl plugin export list

culb:
The documentation lists "HexChat::plugin_pref_del( $setting )" as the method to delete preferences but only plugin_pref_delete is exported.


TingPing:
So.. fix the docs then?

culb:
Maybe a user would prefer to use "plugin_pref_del" instead of "plugin_pref_delete". It's the same length as set and get and could potentially look better in certain code.

TingPing:
Well that could have been discussed when the API was added, I don't think there should be aliases as part of the API to save typing 3 letters.

PR #1943, opened 2017-02-01
fix spelling mistake

EdwardBetts:


TingPing:
Can't rename text events sadly, they are effectively part of our public API used in scripts and themes.

EdwardBetts:
Can we include a comment in textevents.in to document the spelling mistake?

TingPing:
Perhaps, one would have to update `make-te.c` to support comments.

Issue #1948, opened 2017-02-11
Sometimes the scrollbar does not stay at the bottom when returning to a tab


rodri:
This patch, applicable to git master, solves an annoying bug that causes that, when a tab (channel or private) is left with the scrollbar at the bottom, and some new text arrives, and then we select again that tab, sometimes the scrollbar is not completely down and there is hidden text and we have to scroll manually to the bottom.

```diff -Naur hexchat-2.12.4/src/fe-gtk/xtext.c hexchat-2.12.4-fix/src/fe-gtk/xtext.c
--- hexchat-2.12.4/src/fe-gtk/xtext.c	2016-07-05 21:46:48.000000000 +0200
+++ hexchat-2.12.4-fix/src/fe-gtk/xtext.c	2017-02-11 12:35:37.011291906 +0100
@@ -4962,6 +4962,14 @@
 	dontscroll (buf);	/* force scrolling off */
 	xtext->adj->value = buf->old_value;
 	xtext->adj->upper = buf->num_lines;
+
+	/* if the scrollbar was down, keep it down */
+	if (xtext->buffer->scrollbar_down && xtext->adj->value <
+		xtext->adj->upper - xtext->adj->page_size)
+	{
+		xtext->adj->value = xtext->adj->upper - xtext->adj->page_size;
+	}
+


 	if (xtext->adj->upper == 0)
 		xtext->adj->upper = 1;
 	/* sanity check */
```
[scrollbar_down.patch.txt](https://github.com/hexchat/hexchat/files/768484/scrollbar_down.patch.txt)

TingPing:
Ideally submit a PR: https://help.github.com/articles/creating-a-pull-request/

At the very least post a `git` formatted patch with authorship and commit info.

rodri:
Ok. Pull Request created:
https://github.com/hexchat/hexchat/pull/1949

PR #1949, opened 2017-02-11
Always keep the scrollbar at the bottom if it was before

rodri:
This patch hopefully solves a bug that causes that when a tab (channel or private) is left with the scrollbar at the bottom, and some new text arrives, and then we select again that tab, sometimes the scrollbar is not completely down and there is hidden text and we have to scroll manually to the bottom to show all text.
This was previously submitted as an issue in: https://github.com/hexchat/hexchat/issues/1948
Maybe it isn't the ideal way to solve it, but at least I tested it and it seems to work.

### DIFF
--- a/src/fe-gtk/xtext.c
+++ b/src/fe-gtk/xtext.c
@@ -4962,6 +4962,14 @@ gtk_xtext_buffer_show (GtkXText *xtext, xtext_buffer *buf, int render)
 	dontscroll (buf);	/* force scrolling off */
 	xtext->adj->value = buf->old_value;
 	xtext->adj->upper = buf->num_lines;
+
+	/* if the scrollbar was down, keep it down */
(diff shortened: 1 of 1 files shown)


TingPing:
So, xtext is sort of a mess and I'm not too confident in changes to it, somebody who has spent more time than me staring at it mentioned:

```
16:39:01| <leeter> so I know what's going on there... and what that's doing
16:39:13| <leeter> but without fixing all the random ass draws that xtext does I wouldn't accept it
16:40:19| <leeter> that is the fix
16:40:28| <leeter> but they shouldn't be modifying the value directly
```

Somebody really needs to sit down and spend a month making it not trash but that is a bit beyond the scope of this PR. We can leave it open if somebody ever wants to spend time reviewing it.

torhus:
If this fix works, why don't we test it?

Issue #397, opened 2013-02-04
Allow configuration of ctrl+w keyboard shortcut (close channel)


JakobOvrum:
This can be quite annoying as browsers often use the same shortcut to close tabs.

After a short search I wasn't able to find the relevant source, hence this report. I wouldn't mind doing this myself if I was more familiar with the code.

aoloe:
sorry, i don't really see the interest of being able to create an empty channel with ctrl-n and closing the current channel with ctrl-w.

the context menu in the channel list and `/close` are imo more than good enough.
(the first one for those who join channels with the mouse, the second one for those who know about the irc command like `/join`: seriously: i can type faster `/join?  than i can press ctrl-w!).

i would suggest to simply remove both shortcuts (and, once you have gtk-3 you can let people who want the two shortcuts add them...)

adding a confirmation dialog might help... but i'm not sure that it's really needed.

strycore:
I have a patch to disable some keyboard shortcuts for those who want to compile hexchat without those enabled: https://github.com/strycore/buildscripts/blob/master/hexchat/disable-shortcuts.patch

larsks:
I came here to file an RFE for this, but I see everyone already beat me to it.

I'm constantly switching between terminals and hexchat and I type C-w for "backspace word" and then find myself trying to figure out which channel I just closed.  I'll try building a package with the patches from @strycore.

aoloe:
@strycore i'm not sure that disabling ctrl-q is a good idea (for sure, i can live with it!)... but disabling by default ctrl-n and ctrl-w seems the way to go...
would you mind making a pull request with that tiny patch so we can see what the developers think about it?

TingPing:
@aoloe See https://github.com/hexchat/hexchat/issues/397#issuecomment-20154069 for the problem with that.

strycore:
@aoloe I personally don't like Ctrl+q (I have a dedicated extension in Firefox to disable it) but people who are less clumsy than me might sometimes use that shortcut intentionally (unlike me).
I seem to remember that @TingPing didn't want to disable those shortcut and since he's one of the only contributors for one of my projects then I don't want to do anything that might annoy him XD

aoloe:
@TingPing , i agree that ctrl-w and ctr-n are common shortcuts.

but,  personally, i would never expect ctrl-w to quit a channel or ctrl-n to create a new empty channel.
[...]

PR #1952, opened 2017-02-19
Removed ctrl+w keybinding

jakon2048:
Multiple of my friends and I are big HexChat users and we always run into the issue of hitting ctrl+w accidentally (same shortcut as closing a browser window) as we misjudge the current in-focus window. As a result, many annoyances were shed. 

It seems like many other users share this gripe as well: https://github.com/hexchat/hexchat/issues/397

I understand better key-binding optimization may or may not come, so the proposal here is to disable it as a behavior overall. Perhaps when better key-binding customization is supported, users can choose themselves whether or not to enable this shortcut. 

Note this has nothing to do with the ctrl+q behavior.


TingPing:
I suppose this is reasonable as users can add a custom binding in settings for `/close` and it probably isn't overly valuable. Could you add `Closes #397` to the commit message body?

jakon2048:
Thanks. Done.

PR #1953, opened 2017-02-21
Allow the text in the user list hostname column to be changed

culb:
Allow the text in the user list hostname column to be changed


TingPing:
Rebase. Also I'm not sure if just telling people "Modify this undocumented internal list" is exactly a friendly extension point.

culb:
I made a mistake by forgetting to add color. It should be "Allow the text color in the user list hostname column to be changed"

PR #1977, opened 2017-03-17
Fix key exchange and decrypting private messages

eimmot:
Clients will recieve the proper shared secret.

Eliminates error message:
SSL_read: error:05066064:Diffie-Hellman routines:compute_key:no private value

Decrypts private messages with the proper nick, and channel decryption has higher priority.

### DIFF
--- a/plugins/fishlim/plugin_hexchat.c
+++ b/plugins/fishlim/plugin_hexchat.c
@@ -138,7 +138,7 @@ static int handle_outgoing(char *word[], char *word_eol[], void *userdata) {
 static int handle_incoming(char *word[], char *word_eol[], hexchat_event_attrs *attrs, void *userdata) {
     const char *prefix;
     const char *command;
-    const char *recipient;
+    const char *channel = hexchat_get_info(ph, "channel");
     const char *encrypted;
     const char *peice;
     char *sender_nick;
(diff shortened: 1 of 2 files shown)


plugins/fishlim/plugin_hexchat.c
```diff
@@ -138,7 +138,7 @@ static int handle_outgoing(char *word[], char *word_eol[], void *userdata) {
 static int handle_incoming(char *word[], char *word_eol[], hexchat_event_attrs *attrs, void *userdata) {
     const char *prefix;
     const char *command;
-    const char *recipient;
+    const char *channel = hexchat_get_info(ph, "channel");
```
TingPing: I don't believe this value can be relied upon as server hooks happen before a useful context is set and some events like NOTICE have no context.

eimmot: Ya I just used what handle_outgoing was using to get channel name and it worked, I didn't have any problems. There must be a better way to get channel name?

TingPing: Yes, you get it from the message contents like it did before. Perhaps it was grabbing the wrong word or something?

eimmot: Wow, I think my fishlim conf was messed up with self keys, they were being used instead of the senders. I removed the self keys and removed the changes to this file and it works.

Issue #1974, opened 2017-03-14
FiSHLiM Key Exchange not working with other clients


eimmot:
Works fine for hexchat <-> hexchat,
hexchat <-> irssi  key exchange fails, console displays error:

**SSL_read: error:05066064:Diffie-Hellman routines:compute_key:no private value**

The keys seem to go through, but they get corrupted or something. Haven't tested against other clients only irssi

TingPing:
For reference: https://github.com/falsovsky/FiSH-irssi/blob/master/src/DH1080.c

Just a guess would be a base64 issue, as otherwise the files are nearly identical.

PR #1978, opened 2017-03-17
Fix key exchange

eimmot:
Clients will receive the proper shared secret.

Eliminates error message:
SSL_read: error:05066064:Diffie-Hellman routines:compute_key:no private value

### DIFF
--- a/plugins/fishlim/dh1080.c
+++ b/plugins/fishlim/dh1080.c
@@ -189,6 +189,7 @@ dh1080_compute_key (const char *priv_key, const char *pub_key, char **secret_key
 	char *pub_key_data;
 	gsize pub_key_len;
 	BIGNUM *pk;
+	BIGNUM *temp_pub_key = BN_new();
 	DH *dh;
 
   	g_assert (secret_key != NULL);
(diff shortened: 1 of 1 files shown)


TingPing:
This doesn't fix #1974 does it?

eimmot:
#1974 is fixed with this commit, no error appears and both clients have the same shared secret

TingPing:
Cool can you add `Fixes #1974` in the commit message body.

eimmot:
Added!

irgendsontyp:
I'm just curious: Why does setting the public key solve the error indicating that a private key is missing?

eimmot:
On the source of DH_set0_key:
https://github.com/openssl/openssl/blob/0a3dce8257d62cd39a17ff7fdbb1fe4724146f7b/crypto/dh/dh_lib.c#L245

If both dh->pub_key and the input pub_key are null, dh->priv_key is never set (also not set inside fishlim), and that throws the error on compute_key:
https://github.com/openssl/openssl/blob/7928ee4d685b727619555bc1ec0aee805f6fc8c4/crypto/dh/dh_key.c#L165

OpenSSL docs on DH_set0_key:
https://www.openssl.org/docs/man1.1.0/crypto/DH_set0_key.html
> The public and private key values can be set using DH_set0_key(). **The public key must be non-NULL the first time this function is called on a given DH object.** The private key may be NULL. On subsequent calls, either may be NULL, which means the corresponding DH field is left untouched. As for DH_set0_pqg() this function transfers the memory management of the key values to the DH object, and therefore they should not be freed directly after this function has been called.

Issue #1902, opened 2016-12-14
Theme manager relies upon current directory to detect portable-mode


z0rti:
Using: HexChat 2.12.4
hexchat.exe works as expected, using HexChat\config\ folder
thememan.exe complains:
"HexChat installation not found! Check your %APPDATA%/HexChat folder"

TingPing:
How are you launching it?

z0rti:
I launched it with a path "c:\Program Files\HexChat\thememan.exe"
Doubleclick in explorer does work, though...

TingPing:
Yea it relies upon the CWD to find the `portable-mode` file.

PR #1985, opened 2017-04-01
thememan: Check for portable-mode file in the application directory.

learn-more:
Fix for #1902


TingPing:
Please include `Fixes #1902` in the commit message body.

learn-more:
I was not sure if that was allowed, will do in the revised commit.
Need to research why mono doesn't like this.

Arnavion:
Too old to support `?.` probably.

learn-more:
Good point @Arnavion

PR #1986, opened 2017-04-03
Add editorconfig, describing what indenting style to use.

learn-more:
This allows multiple editors (see http://editorconfig.org/ for the full list) to automatically use the correct indenting style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.{c,cpp,h,hpp,m}]
+indent_style = tab
+indent_size = 4
+end_of_line = lf
+
+[*.cs]
+indent_style = space
+indent_size = 4
+end_of_line = lf


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.c]
```
TingPing: plus header files and c++ files.

TingPing: And probably should indicate tab size of 4

learn-more: For space indent, yes.
For tab indent it does not matter, everyone has their own preference?
(tabs are tabs)

I have [*.{c,cpp,h,hpp}] now, any more extensions I should add?

TingPing: .m

The codebase is fairly inconsistent using tabs for alignment, thus their size tends to matter.

learn-more: In that case it also makes sense to include the newline style?

TingPing: This is what I had in another branch: https://github.com/hexchat/hexchat/blob/wip/meson/.editorconfig

learn-more:
These were the files relevant for me, and suggested combined.
Let's add new extensions on demand?

Issue #1999, opened 2017-05-06
HexChat won't start, no error message.


deadalnix:
HexChat won't start on my win10 machine. I get no error message or anything, nothing happens. Trying to launch HexChat via command line do not prompt anything on the standard output.

I tried to uninstall and reinstall HexChat, and to remove the %APPDATA%\HexChat folder, but nothing seems to work.

I'm not sure what is going on, or even how to get any information about what is going on.

TingPing:
Does the "Safe mode" hexchat launcher work?

deadalnix:
Hi @TingPing .

The same thing happen when launching the safe mode. And by the same thing, i mean literally nothing, the program quit immediately, no error message, no window opened, nothing.

deadalnix:
Ok I have a crash dump file. Who should I send that to ?

learn-more:
Are you able to attach it here?

deadalnix:
I have no idea how to attach files to github issues. Is that a thing ?

TingPing:
> Attach files by dragging & dropping, , or pasting from the clipboard.

deadalnix:
[hexchat.crashdump.zip](https://github.com/hexchat/hexchat/files/988017/hexchat.crashdump.zip)

learn-more:
It's an assert in cairo.dll,
@TingPing : are there symbol files for releases somewhere?

TingPing:
> @TingPing : are there symbol files for releases somewhere?

@tomek: are there symbol files for releases somewhere?

tomek:
https://dl.hexchat.net/hexchat/pdbs/

deadalnix:
Hi everybody.

After a round of windows update, HexChat is back to work. It looks like the problem was on windows's side and not on yours.

learn-more:
It was this assertion:
https://bugs.launchpad.net/ubuntu/+source/cairo/+bug/1476736

ProjectRevoTPP:
Having this issue after the Windows Creators Update. Safe mode doesn't work, tried a reinstall of hex chat, same thing.

chibill:
I am having the same problem after a windows update... Reinstalling does not seem to fix the problem.

parkrrr:
I have the same behavior on Windows 10, but mine chokes on python:
```ntdll!NtTerminateProcess+0x14
ntdll!RtlExitUserProcess+0xbf
KERNEL32!ExitProcess+0xa
[...]

PR #2011, opened 2017-05-28
servlist: skip empty lines.

learn-more:
Should fix out of range from #1999


TingPing:
Can put `Fixes #1999` in the commit message.

learn-more:
I didn't do that, since this fixes a report from someone who is not the original reporter,
and we can't be sure that this is the same problem.

Issue #1856, opened 2016-10-24
Patch for new sound setting


dstasinski:
I've created a patch that may interest the team or others.  I make a lot of use of event sounds but there are times when I want them turned off so I added:

  /set SOUNDS ON|OFF

I've documented the patch files at http://www.saidsimple.com/daniel/blog/117915/temporarily-disable-sounds-in-hexchat

Issue #668, opened 2013-07-03
HexChat doesn't show in volume mixer


soulsynapse:
I like where my system volume is and hexchat's volume is mindnumbingly loud, especially for beeps and such. It seems there is no slider and the program doesn't show up in windows mixer, is the feature already there or am I just not seeing it?

Atario:
Seems this could be solved pretty easily by playing a fractional second of silence on startup…

PR #2012, opened 2017-06-06
Checking the "Omit alert when marked as being away" box disables sounds when marked as being away.

YuGiOhJCJ:
Sometimes, I want to disable my sound notifications because I am busy so I go to Settings > Preferences > Sounds then I change the value of each item.
Of course, it is not fast and easy to do.

If you think that I could fix this problem by muting hexchat in the operating system through the mixer, it is not possible.
Indeed, my mixer only shows applications that are currently using audio so hexchat is not in the list of applications.

So, I fix this problem by adding an "Enable sounds" setting in the Settings > Preferences > Sounds menu.

An "Enable sounds" setting

This fix is related to these issues (but it does not fix them):
https://github.com/hexchat/hexchat/issues/1856
https://github.com/hexchat/hexchat/issues/668

### DIFF
--- a/src/common/cfgfiles.c
+++ b/src/common/cfgfiles.c
@@ -547,8 +547,6 @@ const struct prefs vars[] =
 
 	{"perl_warnings", P_OFFINT (hex_perl_warnings), TYPE_BOOL},
 
-	{"sounds", P_OFFINT (hex_sounds), TYPE_BOOL},
-
 	{"stamp_log", P_OFFINT (hex_stamp_log), TYPE_BOOL},
 	{"stamp_log_format", P_OFFSET (hex_stamp_log_format), TYPE_STR},
 	{"stamp_text", P_OFFINT (hex_stamp_text), TYPE_BOOL},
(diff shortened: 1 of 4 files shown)


TingPing:
You already have *"Omit alerts when marked as being away"*, so I think that should just be re-used instead of a new option.

YuGiOhJCJ:
I did a test by checking the box in Settings > Preferences > Alerts > Omit alerts when marked as being away, then I typed "/away" in a channel to be marked as being away.

In the same IRC network and channel but with an other client and an other name I sent a message to the user connected through the hexchat client.
I heard a sound.

So, it means that the sounds (in Settings > Preferences > Sounds) are not impacted by the alert settings (in Settings > Preferences > Alerts).

That's why I still need a new option.

TingPing:
> So, it means that the sounds (in Settings > Preferences > Sounds) are not impacted by the alert settings (in Settings > Preferences > Alerts).

Correct, and that is why I said to make it respect that option.

YuGiOhJCJ:
Oh OK I got it.
Now, checking the box ```Omit alert when marked as being away``` box and typing ```/allserv away``` does exactly what I want: I don't heard a sound anymore.

However, as Alerts and Sounds are in a separate setting page, I thought that it was a good idea that Alerts do not impact Sounds.
That's why I would prefer a new option in the Sounds setting page.
But the solution that you propose is a good compromise, so that is OK for me.

PR #2018, opened 2017-06-25
rename normal status icon to tray_none.png

### LINKED ISSUE #2017: Fix issue with icons
Currently, four icons are used by the program:

1. `hexchat.png`
2. `tray_fileoffer.png`
3. `tray_highlight.png`
4. `tray_message.png`

All four of the icons are used in the tray, but `hexchat.png` is also used elsewhere (eg: in the alt+tab pane and the windows list in the panel).

Therefore, when one uses the `hardcode-tray` script in order to hardcode the indicator icon, changing `hexchat.png` changes the icons for other places too, spoiling the fun. Perhaps, if you used a separate icon for the panel and rest of the places, it would be better to use hexchat in conjunction with the hardcode-tray script.

bil-elmoussaoui:
Should fix #2017 
The icon name is not that good, i can rename it to whatever you guys want :+1: 
I've tried the fix and everything works as expected


TingPing:
You don't need to actually make a copy of the icon in-tree since its the same as `hexchat.png`.

I think `tray_normal` might be a better name than `none`.

bil-elmoussaoui:
Makes sense, i will update my commit this evening

bil-elmoussaoui:
@TingPing Should be fixed now

PR #2035, opened 2017-08-06
Adjust use of ellipsis (...) in menus according to Gnome HIG

torhus:
https://developer.gnome.org/hig/stable/writing-style.html.en

I kept it clean by not also tweaking wording and capitalization, not sure if I should do that in a seperate PR or update this one.

### DIFF
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File…"), "gtk-floppy",
+		_("_Send a Fileâ€¦"), "gtk-floppy",
 		_("_User Info (WhoIs)"), "gtk-info",
-		_("_Add to Friends List…"), "gtk-add",
+		_("_Add to Friends Listâ€¦"), "gtk-add",
 		_("_Ignore"), "gtk-stop",
 		_("O_perator Actions"),
 
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File..."), "gtk-floppy",
+		_("_Send a File…"), "gtk-floppy",
 		_("_User Info (WhoIs)"), "gtk-info",
-		_("_Add to Friends List..."), "gtk-add",
+		_("_Add to Friends List…"), "gtk-add",
 		_("_Ignore"), "gtk-stop",
 		_("O_perator Actions"),
 
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File…"), "gtk-floppy",
+		_("_Send a File" ELLIPSIS), "gtk-floppy",
 		_("_User Info (WhoIs)"), "gtk-info",
-		_("_Add to Friends List…"), "gtk-add",
+		_("_Add to Friends List" ELLIPSIS), "gtk-add",
 		_("_Ignore"), "gtk-stop",
 		_("O_perator Actions"),
 
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -1,4 +1,4 @@
-﻿/* X-Chat
+/* X-Chat
  * Copyright (C) 1998 Peter Zelezny.
  *
  * This program is free software; you can redistribute it and/or modify
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -1,4 +1,4 @@
-/* X-Chat
+﻿/* X-Chat
  * Copyright (C) 1998 Peter Zelezny.
  *
  * This program is free software; you can redistribute it and/or modify
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File" ELLIPSIS), "gtk-floppy",
+		_("_Send a File…"), "gtk-floppy",
 		_("_User Info (WhoIs)"), "gtk-info",
-		_("_Add to Friends List" ELLIPSIS), "gtk-add",
+		_("_Add to Friends List…"), "gtk-add",
 		_("_Ignore"), "gtk-stop",
 		_("O_perator Actions"),
 
--- a/src/common/hexchat.c
+++ b/src/common/hexchat.c
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File"), "gtk-floppy",
+		_("_Send a File..."), "gtk-floppy",
 		_("_User Info (WhoIs)"), "gtk-info",
-		_("_Add to Friends List"), "gtk-add",
+		_("_Add to Friends List..."), "gtk-add",
 		_("_Ignore"), "gtk-stop",
 		_("O_perator Actions"),
 
--- a/src/common/common.vcxproj
+++ b/src/common/common.vcxproj
@@ -118,11 +118,5 @@ powershell -File "$(SolutionDir)..\win32\version-template.ps1" "$(SolutionDir)..
 
       ]]></Command>
     </PreBuildEvent>
-    <ClCompile>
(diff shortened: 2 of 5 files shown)


src/common/hexchat.c
```diff
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File"), "gtk-floppy",
+		_("_Send a File..."), "gtk-floppy",
```
TingPing: These should be the `…` character not 3 periods.

torhus: Ok, even though it's independent characters everywhere else? Should I also change the others?

torhus: I'm also not sure how this relates to the translations, the PO files.

TingPing: Yes please change all of them. Yes everything in HexChat is UTF-8.

torhus: VS 2017 displays ellipsis characters just fine when using UTF-8, hopefully the compiler will accept them too. In the Windows UI font (Segoe UI), the character looks exactly the same as just three periods, should probably check that it doesn't look squashed on other platforms.

src/common/hexchat.c
```diff
@@ -883,9 +883,9 @@ xchat_init (void)
 		"NAME ENDSUB\n"				"CMD \n\n",
 
 		_("_Open Dialog Window"), "gtk-go-up",
-		_("_Send a File..."), "gtk-floppy",
+		_("_Send a File�"), "gtk-floppy",
```
TingPing: Either github is breaking rendering or this is the wrong character?

torhus:
It didn't work on Windows, the ellipsis was displayed as three random characters. So maybe GTK+ is interpreting narrow strings as ANSI (often Windows-1252) on Windows?  Escape sequences won't help then...

TingPing:
We already have unicode characters with escape sequences: https://github.com/hexchat/hexchat/blob/master/src/common/text.c#L54

torhus:
Unicode replacement character, I suppose that's not visible anywhere? What I get in the menus are "â€¦", which corresponds exactly to the UTF-8 sequence for the ellipsis, interpreted in Windows-1252. I don't know if there is a way to make GTK+ handle this correctly, not familiar with it. Why not just stick with the old solution to avoid all the headache?

torhus:
Ok, I didn't see the second part of your comment. I think the the UTF-8 is ok as far as MSVC is concerned, but either GTK is using the 8-bit Windows API and just handing the unaltered strings to it, or converting to UTF-16 under the assumption that the strings are Windows-1252. In either case, the result should be what I see.

Arnavion:
`"\342\200\246"`

torhus:
Doesn't help if GTK doesn't convert it correctly, the Windows API doesn't speak UTF-8. The result would be the same. I didn't find anything useful when googling the issue either.

TingPing:
There isn't a problem in GTK.

torhus:
How come the UTF-8 three code units show up in the menu as three characters from Windows-1252 then? I dont' have any other ideas. Do you happen to know where the source of GTK's Windows implementation is?

TingPing:
> How come the three UTF-8 code units show up in the menu as three characters from Windows-1252 then, after passing through GTK?

This is probably related to how the compiler is dealing with the strings rather than anything else. All strings in GTK are UTF-8 and any interaction with Windows does convert to UTF-16 for those APIs.

> Do you happen to know where the source of GTK's Windows implementation is?

https://github.com/GNOME/gtk/tree/gtk-2-24/gdk/win32

torhus:
Is it ok to abuse the PR to make build for testing something..?

TingPing:
Feel free.

torhus:
I tried encoding the file as Windows-1252, and adding five characters outside of ASCII to see what would happen. They all showed up in the menu. Which means that there is no conversion from UTF-8 to UTF-16 begin done, it's either from Windows-1252 to UTF-16, or none at all.

I tried adding some UTF-8 through escape sequences, but that was dropped somewhere along the line. Maybe by the compiler, maybe by a conversion function. But encoding the file as UTF-8 works, it's just the conversion that's wrong.

The g_get_charset functions returns Windows-1252 on Windows, and I see it's used in some conversion code, maybe that's an issue? Not familiar enough with the code to understand what is going on here yet.

Arnavion:
>I tried encoding the file as Windows-1252, and adding five non-ASCII characters to see what would happen. They all showed up correctly in the menu. Which means that there is no conversion from UTF-8 to UTF-16 begin done, it's either from Windows-1252 to UTF-16, or none at all. Of course, there could be two conversions, the other being (incorrectly) from 1252 to UTF-8.

GTK expects you to give utf-8 strings. If you give it non-utf8 strings there's no defined behavior what happens. In this case it appears to use them verbotem, but it doesn't matter.

>I tried adding some UTF-8 through escape sequences, but that was dropped somewhere along the line.

Assuming you're talking about [this](https://github.com/hexchat/hexchat/pull/2035/commits/46bf9e0a6e796e5331fb20a017ece7a400759350#diff-60a599507a4c698f52691b6134ca81e3R1808) you have a typo - `\a6` instead of `\xa6`.

Your current code that uses octal sequences is correct.

image

torhus:
Well, this was confusing. If it's really the compiler doing this, maybe we should try adding the /utf-8 switch to its command line? But it works this way too, I just added an ELLIPSIS macro with the escape sequence string. Someone else will have to check if it works on other platforms.

TingPing:
If there is a cleaner way than macros that would be nice, did you try the BOM method @Arnavion mentioned?

Arnavion:
I don't think you should add a BOM.

TingPing:
@Arnavion Reason being?

torhus:
No problems so far...

Arnavion:
>@Arnavion Reason being?

Just idealistic. It's a Windows-ism.

torhus:
I have never used it before, but I think having an invisible encoding identifier at the beginning of the file makes sense. At least when the encoding is already complex.

TingPing:
> I think the best solution might be to add the compiler switch and see if that solves it, so we don't risk breaking anyone's tools on other platforms.

Sounds good.

torhus:
Any opinions on this, is it ready to be merged?

TingPing:
> GTK+ headers all contain the line

>> #pragma execution_character_set("utf-8")

> which makes the VS 2015 compiler fail with error C3437 when trying to specify the encoding with a compiler flag:

So can we not use that?

Arnavion:
build.ps1 adds it. If you want to use `/utf-8` here you also have to make every gtk-win32 vcxproj use it.

torhus:
This was supposed to be about the use of ellipsis in menus, maybe the more complicated UTF-8 question should be separate? This isn't what I signed up for anymore.

TingPing:
Well you don't get ellipsis without dealing with the encoding problem.

torhus:
What do you mean, I mostly removed them...

TingPing:
Via using BOM atm right? Well as mentioned there might be better ways...

torhus:
The original ellipses were just three commas. It looks the same, at least on Windows.

But if you remove the pragma from the Gtk headers, we could probably use the compiler flag. I think the pragma is the wrong solution, unless there was a technical reason for using them that I am not aware of.

torhus:
Ok, I think I've got it. But I didn't try to build GTK+.

- Remove #pragma execution_character_set("utf-8") from all GTK headers
- Convert atkversion.h to UTF-8
- Add /utf-8 build flag to common and fe-gtk projects

That's it. Are the VS project files generated by a tool, or should I just fix them? Would be great if someone else could do the GTK part, I just downloaded the binaries.

TingPing:
For now just edit them directly but maybe @fanc999 has some thoughts about changing this upstream?

torhus:
Right, are you going to fix the GTK build, then? No point otherwise.

TingPing:
```
22:24:27| <TingPing> shouldn't upstream files still be specifying encoding?
22:24:52| <Arnavion> No, because they assume utf-8, and it works for the most part because they only use non-ascii in comments
22:25:21| <Arnavion> So on en-us windows, cl.exe doesn't complain. On ja-ja windows (like mine), it does complain with C4819
22:26:04| <Arnavion> If you absolutely want to go the /utf-8 route, then as I said in the comment, remove the pragma-adding code from build.ps1, put /utf-8 in stack.props, and it should probably work
22:27:36| <Arnavion> As for upstreaming it, /utf-8 is new in VS2015, so they have to make sure to not add it to the vcxproj for earlier VS
```

So I think we can just deal with this locally for now. Anyway sorry if this went way beyond the scope you wanted but sometimes that happens...

torhus:
How do we deal with it locally? Can you fix the GTK part so the build works?

TingPing:
Ok so discussing it with @Arnavion.. lets just go back to the ugly macro solution we started with. It requires no modifications to our builds, it doesn't use any MS specific BOM, and it should work fine even if a bit ugly.

torhus:
Ok. I suppose you'll do a squash and merge in this case...

PR #2042, opened 2017-08-19
UI text consistency, capitalization etc.

torhus:
I've probably forgotten something, please let me know.

### DIFF
--- a/src/fe-gtk/setup.c
+++ b/src/fe-gtk/setup.c
@@ -161,11 +161,11 @@ static const setting appearance_settings[] =
 	{ST_EFILE, N_ ("Background image:"), P_OFFSETNL (hex_text_background), 0, 0, sizeof prefs.hex_text_background},
 
 	{ST_HEADER, N_("Transparency Settings"), 0,0,0},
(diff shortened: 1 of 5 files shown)


src/fe-gtk/setup.c
```diff
@@ -556,14 +556,14 @@ static const setting advanced_settings[] =
 	{ST_HEADER,	N_("Auto Copy Behavior"),0,0,0},
 	{ST_TOGGLE, N_("Automatically copy selected text"), P_OFFINTNL(hex_text_autocopy_text),
 					N_("Copy selected text to clipboard when left mouse button is released. "
-						"Otherwise, CONTROL-SHIFT-C will copy the "
+						"Otherwise, Ctrl+Shift+C will copy the "
```
TingPing: I think we should use `Control`.

torhus: Ok, any particular reason? I just followed this: https://developer.gnome.org/hig/stable/keyboard-input.html.en

TingPing: Hmm, nevermind you are right and it matches our keyboard shortcuts dialog.

torhus: Cool :)

PR #2055, opened 2017-09-15
Change windows from starting with "Hexchat: " to ending with " - Hexchat"

rchasepatterson:
This changes all Hexchat windows from starting with "Hexchat: " to ending instead with " - Hexchat"

This is done because particularly when Hexchat is used such that all buffers open new windows, it becomes impossible to distinguish any window from another in the window manager window list.  By moving the program branding to the end of the string, the possibly truncated window title visible will provide the most useful information to the user first.

This also makes Hexchat behave more like most other programs (Firefox, Chrome, Thunderbird, etc.) where the branding is at the end of the title.  (Presumably for the same reason.) 

This pull request touches a lot of files, most of them are .po files for the translations as the ": " was a part of the msgid and so each entry needed to be changed.  No change in the style of these messages has been made, and so now the msgid ends with " - ".  Additionally, a number of files using Unicode had a mixture of double-byte and single-byte colons "：" and ":", all instances have been replaced with single-byte at this time.

I personally also believe the "Dialog with" text in the window title for query buffers should be done away with, as it gets in the way of seeing the nick as above when titles are truncated, but this pull request does not make that change.

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -3238,7 +3238,7 @@ msgstr ""
 
 #: ../src/fe-gtk/banlist.c:808
 #, c-format
-msgid ": Ban List (%s)"
+msgid "Ban List (%s) - "
 msgstr ""
 
 #: ../src/fe-gtk/banlist.c:843 ../src/fe-gtk/notifygui.c:427
@@ -3277,7 +3277,7 @@ msgstr ""
 
 #: ../src/fe-gtk/chanlist.c:720
 #, c-format
-msgid ": Channel List (%s)"
+msgid "Channel List (%s) - "
 msgstr ""
 
 #: ../src/fe-gtk/chanlist.c:794
@@ -3369,7 +3369,7 @@ msgid "Cannot resume the same file from two people."
 msgstr ""
 
 #: ../src/fe-gtk/dccgui.c:802
-msgid ": Uploads and Downloads"
+msgid "Uploads and Downloads - "
 msgstr ""
 
 #: ../src/fe-gtk/dccgui.c:819 ../src/fe-gtk/dccgui.c:1060
@@ -3427,7 +3427,7 @@ msgid "Open Folder..."
 msgstr ""
 
 #: ../src/fe-gtk/dccgui.c:1049
-msgid ": DCC Chat List"
+msgid "DCC Chat List - "
 msgstr ""
 
 #: ../src/fe-gtk/dccgui.c:1062
@@ -3624,7 +3624,7 @@ msgid "Select a row to get help information on its Action."
 msgstr ""
 
 #: ../src/fe-gtk/fkeys.c:811
-msgid ": Keyboard Shortcuts"
+msgid "Keyboard Shortcuts - "
 msgstr ""
 
 #: ../src/fe-gtk/gtkutil.c:126
@@ -3669,7 +3669,7 @@ msgid "Enter mask to ignore:"
 msgstr ""
 
 #: ../src/fe-gtk/ignoregui.c:350
-msgid ": Ignore list"
+msgid "Ignore list - "
 msgstr ""
 
 #: ../src/fe-gtk/ignoregui.c:358
@@ -3705,7 +3705,7 @@ msgid "Channel name too short, try again."
 msgstr ""
 
 #: ../src/fe-gtk/joind.c:133
-msgid ": Connection Complete"
+msgid "Connection Complete - "
 msgstr ""
 
 #: ../src/fe-gtk/joind.c:161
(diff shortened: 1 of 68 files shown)


TingPing:
> This pull request touches a lot of files, most of them are .po files for the translations as the ": " was a part of the msgid and so each entry needed to be changed. No change in the style of these messages has been made, and so now the msgid ends with " - ". Additionally, a number of files using Unicode had a mixture of double-byte and single-byte colons "：" and ":", all instances have been replaced with single-byte at this time.

If we are going this far lets actually fix these translations. The window title and `HexChat` should probably be translated separately. I also wonder how to handle this in RTL setups, should it be `Hexchat -Foo`?

rchasepatterson:
The `HexChat` in the title comes from a defined value `DISPLAY_NAME` in `fe-gtk.h` which is used only in a handful of locations outside of the window titles that have been changed in this PR.

If we want to replace the reference to `DISPLAY_NAME` to something else, there already exists a `msgid "HexChat"` which can be used for this purpose that seems to have a few localized names.

Looking at screenshots of arabic language OSs it seems that the truncation of window titles is inconsistent from what I can find.  However if we assume RTL window naming and truncation on the left instead of the right, even without placing `" - HexChat"` at the end HexChat already produces output that places the nick and network name on the right by way of its ordering of arguments.

This can be fixed in the individual .po files for the languages needing them by providing the position of the argument in the translated msgstr in the form of `%n$` as per https://www.gnu.org/software/gettext/manual/gettext.html#c_002dformat

What changes would you ideally like to see before merging this?

TingPing:
All I would like is for the `HexChat` and associated `-` to translated separately from the titles if possible. I would think that would lead to the best translations.

rchasepatterson:
Everywhere `DISPLAY_NAME` was being used hard-coded in the window title (and the about window) has been changed to instead use a localized version of it, if one exists.

Additionally, translations have been changed from simply ending in ` - ` to ending in ` - %s` which means that the .po file can be used to adjust if the title should show at the front, end, or anywhere else, within the title.

For example, from sr.po:
```
#: ../src/fe-gtk/plugin-tray.c:655
#, c-format
msgid "%u channel messages. - %s"
msgstr "%u порука на каналу. - %s"
```

May produce `1 порука на каналу. - ХексЧет`

If we instead wanted it to show the title at the front, RTL style, the translation would be changed accordingly:

```
#: ../src/fe-gtk/plugin-tray.c:655
#, c-format
msgid "%u channel messages. - %s"
msgstr "%2$s - %1$u порука на каналу."
```

Which would instead produce `ХексЧет - 1 порука на каналу.`

TingPing:
> If we instead wanted it to show the title at the front, RTL style, the translation would be changed accordingly

My assumption is nearly no translators know that format and to use it they would have to modify dozens of lines. I probably don't know a better solution without allocating a newly formatted string everywhere though.

rchasepatterson:
I've checked with a friend of mine who is fluent in Arabic, and he has confirmed that in RTL usage that the title of the program tends to be on the left, and that text is cut off starting on the left.

> My assumption is nearly no translators know that format and to use it they would have to modify dozens of lines. I probably don't know a better solution without allocating a newly formatted string everywhere though.

I don't know a better solution either, this seemed the simplest while fitting with the gettext documentation for how arguments should be reordered.  As their documentation states: 

> Although format strings with positions that reorder arguments [...] are a POSIX/XSI feature and not specified by ISO C 99, translators can rely on this reordering ability: On the few platforms where printf(), fprintf() etc. don’t support this feature natively, libintl.a or libintl.so provides replacement functions, and GNU <libintl.h> activates these replacement functions automatically. 

So hopefully anyone wanting to translate into a right-to-left language would already be aware of this.  Failing that, there are only a small number of right to left languages (none of which HexChat has localization for at current) and it may be simplest to make that change there after translation work is done, since it is only the title that needs be moved.

TingPing:
Alright could you rebase these and remove all modifications to `.po` files?

rchasepatterson:
What would you like me to rebase it to?  I see my branch as two commits ahead of the upstream master branch.  I can combine it into one commit if that's what you'd like.

Also the changes to the .po files are necessary for them to display the titles correctly, without them I do not believe the translations will work and it'll default back to English.  This is because the `msgid` has changed to remove the leading `: ` and has been replaced with a trailing ` - %s` (which is what allows it to be changed in the `msgstr` as well).

TingPing:
The reason being all translations are handled on transifex so after this is merged I upload the .pot file there and then get the .po files back.

I suppose I can upload your po files at the same time...

rchasepatterson:
That makes sense, I wasn't aware that that was the case.  I'll keep that in mind for any future changes to the translations.

I see you've merged it, is there anything else you need from me?

TingPing:
Nope.

PR #2056, opened 2017-09-17
Spelling: SOCKS5, Byte per second

comradekingu:
As per https://en.wikipedia.org/wiki/SOCKS


TingPing:
This is also spelled incorrect in `setup.c` I believe.

comradekingu:
@TingPing Looked a bit like it would start breaking things. Also changed to "Byte per second" for referencing the unit of measure.

TingPing:
Is `bytes per seconds` actually wrong? It sounds better to me and I see it widely in use.

comradekingu:
@TingPing I think one is the unit of measure and the other is actual use when it is beyond 1.

PR #2058, opened 2017-09-18
Fix rfc_{n,}casecmp functions

josephbisch:
Implementation of rfc_casecmp sourced from TingPing.

### DIFF
--- a/src/common/util.c
+++ b/src/common/util.c
@@ -1018,53 +1018,39 @@ make_ping_time (void)
 	return (timev.tv_sec - 50000) * 1000 + timev.tv_usec/1000;
 }
 
-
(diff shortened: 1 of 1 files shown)


src/common/util.c
```diff
@@ -1018,53 +1018,39 @@ make_ping_time (void)
...
+			return (c1 - c2);
+		n--;
 		s1++;
 		s2++;
-		n--;
-		if (n == 0 || (*s1 == '\0' && *s2 == '\0'))
-			return 0;
 	}
-	return (res);
+	return 0;
```
TingPing: If one string is shorter than the other and end before `n` it still returns `0`.

PR #2064, opened 2017-10-02
Allow chanlist_minusers to be 0 to show empty channels

OndrejBakan:
Hello,

I think it should be allowed to show channels with 0 users in /list. I tried to make this change and it is working for me, but I think it could be in the official release too.

### DIFF
--- a/src/fe-gtk/chanlist.c
+++ b/src/fe-gtk/chanlist.c
@@ -310,8 +310,8 @@ chanlist_do_refresh (server *serv)
 	else
 	{
 		/* download all, filter minusers locally only */
-		serv->p_list_channels (serv, "", 1);
-		serv->gui->chanlist_minusers_downloaded = 1;
+		serv->p_list_channels (serv, "", 0);
+		serv->gui->chanlist_minusers_downloaded = 0;
 	}
 
 /*	gtk_spin_button_set_range ((GtkSpinButton *)serv->gui->chanlist_min_spin,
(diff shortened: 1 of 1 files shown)


src/fe-gtk/chanlist.c
```diff
@@ -310,8 +310,8 @@ chanlist_do_refresh (server *serv)
 	else
 	{
 		/* download all, filter minusers locally only */
-		serv->p_list_channels (serv, "", 1);
-		serv->gui->chanlist_minusers_downloaded = 1;
+		serv->p_list_channels (serv, "", 0);
```
TingPing: The idea seems fine but I'm not sure about this line. I never want us to actually request that many potential channels unless the *user* explicitly does so.

src/fe-gtk/chanlist.c
```diff
@@ -727,7 +727,7 @@ chanlist_opengui (server *serv, int do_refresh)
 
 	if (!serv->gui->chanlist_minusers)
```
TingPing: This can be 0 now...

TingPing: 0 is a valid value for this now.

src/fe-gtk/chanlist.c
OndrejBakan: That's fine. My IRC network is some kind of bridged network, so it works for me, but I understand it probably won't work for real IRC networks with tons of channels.

OndrejBakan:
@TingPing Sorry to bother you, but after compiling (I am really noob at it, I just edited the files and then ran build in VS) Ctrl+W shortcut does not work. It worked until that change. Any idea why?

TingPing:
> Ctrl+W shortcut does not work. Any idea why?

That shortcut was disabled months ago, you can add a custom one in *Settings > Keyboard Shortcuts* that calls `/close`.

PR #2065, opened 2017-10-02
Use g_utf8_make_valid if available for cleaner utf8 handling

josephbisch:


src/common/text.c
```diff
@@ -814,13 +814,29 @@ text_convert_invalid (const gchar* text, gssize len, GIConv converter, const gch
...
 text_fixup_invalid_utf8 (const gchar* text, gssize len, gsize *len_out)
 {
+#if GLIB_CHECK_VERSION (2, 52, 0)
+	gchar *result = g_utf8_make_valid (text, len);
+	if (len_out) {
+		/* If g_utf8_make_valid had to make a change, then we can use strlen
+		 * because result is definitely terminated. But if the text was already
+		 * valid, then result may not be terminated.
+		 */
+		if (g_utf8_validate (text, len, NULL)) {
```
TingPing: This will always be slower than `strlen()` so just call `strlen()`.

src/common/text.c
```diff
@@ -814,13 +814,29 @@ text_convert_invalid (const gchar* text, gssize len, GIConv converter, const gch
 gchar *
 text_fixup_invalid_utf8 (const gchar* text, gssize len, gsize *len_out)
 {
+#if GLIB_CHECK_VERSION (2, 52, 0)
+	gchar *result = g_utf8_make_valid (text, len);
+	if (len_out) {
```
TingPing: Braces on newlines.

josephbisch:
Let me know if I should squash the commits.

src/common/text.c
```diff
@@ -814,13 +814,22 @@ text_convert_invalid (const gchar* text, gssize len, GIConv converter, const gch
 gchar *
 text_fixup_invalid_utf8 (const gchar* text, gssize len, gsize *len_out)
 {
+#if GLIB_CHECK_VERSION (2, 52, 0)
+	gchar *result = g_utf8_make_valid (text, len);
```
TingPing: Since we try to target an older glib you have to add `G_GNUC_BEGIN_IGNORE_DEPRECATIONS`/`G_GNUC_END_IGNORE_DEPRECATIONS` around this function to silence complaints.

Issue #2069, opened 2017-10-06
No option to disable rejoining channels on auto reconnect


aucampia:
When using hexchat through ZNC from multiple computers there is a situation where, if one computer is suspended with hexchat open and connected to ZNC and another computer leaves a channel then, the channel will be rejoined when the suspended computer is un-suspended.

The reason for this is that hexchat is not aware that the channel was parted and automatically rejoins to all open channels on reconnect.

Ideally there should be an option to disable auto rejoining on reconnect to better accomodate situations like this.

This problem can be simulated as follow:

### 1. Setup two hexchat instances

```
hexchat --cfgdir="${HOME}/.config/hexchat.a"
hexchat --cfgdir="${HOME}/.config/hexchat.b"
```

**NOTE:**

 - Connect each instance only to ZNC and nothing else as the commands rely on only 1 tcp connect
 - Make sure logging is configured as pid is obtained from log files
 - Join some channels (e.g. `###sandbox` and `####sandbox`)

### 2. Stop instance A and kill it's single TCP connection

```
## stop 
kill -SIGSTOP $(find "${HOME}/.config/hexchat.a/logs/" -print0 | xargs -0 fuser 2>&1 | gawk 'match($0,/^.*:\s+(\S+)$/,captured){ print captured[1]; exit 0 }')

## kill tcp connection (you can use ctrl+c once you see line ending in win 0)
lsof  -n -P -p $(find "${HOME}/.config/hexchat.a/logs/" -print0 | xargs -0 fuser 2>&1 | gawk 'match($0,/^.*:\s+(\S+)$/,captured){ print captured[1]; exit 0 }') 2>/dev/null | gawk '( $8 == "TCP" ){ print gensub("^([^:]+):([0-9]+)->([^:]+):([0-9]+)$", "src host \\1 and src port \\2 and dst host \\3 and dst port \\4", "g", $9) }' | tr '\n' '\000' | xargs -0 -t bash -c 'sudo tcpkill -1 ${1}' /dev/null
```

### 3. From instance B close `####sandbox` from instance B

```
/part ####sandbox going to test now
```

### 4. Resume instance A

```
kill -SIGCONT $(find "${HOME}/.config/hexchat.a/logs/" -print0 | xargs -0 fuser 2>&1 | gawk 'match($0,/^.*:\s+(\S+)$/,captured){ print captured[1]; exit 0 }')
```

aucampia:
To test the change:

### 1. Setup two hexchat instances

```
[...]

PR #2070, opened 2017-10-06
Add an option to disable rejoin on auto reconnect

aucampia:
This change adds an option `irc_reconnect_rejoin` (default ON)
which when turned OFF will prevent hexchat from rejoining channels
after an auto reconnect.

hexchat/hexchat#2069


aucampia:
For testing instructions see https://github.com/hexchat/hexchat/issues/2069#issuecomment-334895880

PR #2072, opened 2017-10-10
Added BuddyIM and updated ElectroCode to SSL

IotaSpencer:


src/common/servlist.c
```diff
@@ -138,7 +139,7 @@ static const struct defaultserver def[] =
 	{0,			"irc.inet.tele.dk"},
 
 	{"ElectroCode", 0, 0, 0, LOGIN_SASL, 0, TRUE},
-	{0,			"irc.electrocode.net"},
+	{0,			"irc.electrocode.net/+6697"},
```
TingPing: This already used ssl, thats what the last TRUE is above. Same should apply to buddyim.

src/common/servlist.c
```diff
@@ -139,6 +140,7 @@ static const struct defaultserver def[] =
 	{0,			"irc.inet.tele.dk"},
 
 	{"ElectroCode", 0, 0, 0, LOGIN_SASL, 0, TRUE},
+	{0,			"irc.electrocode.net/6667"},
```
TingPing: No.. Remove the ports from both of your entries and remove the duplicated servers.

PR #2081, opened 2017-10-17
Do not set opacity when it's disabled

### LINKED ISSUE #2077: Windows: Transparency problem - would like to disable alpha blending (WS_EX_LAYERED)
Hi,

my problem: Because of the activated alpha blending, HexChat is entirely invisible in my VNC session.

The long story: I often use VNC with a "Screen Blanking" feature which uses alpha blending to work. It would be very nice to have an option to disable transparency completely. This means that the application does not set WS_EX_LAYERED. Currently we can set the "Window Opacity" in the "Preferences" dialog, but increasing this value to 255 doesn't mean that alpha blending is not used/deactivated. I would be happy if there is a checkbox which controls this behaviour (switching it on or off) or that alpha blending will be disabled when the opacity value is set to 255.

Thanks for reading this and best wishes,

Christian

learn-more:
Should fix #2077

### DIFF
--- a/src/fe-gtk/maingui.c
+++ b/src/fe-gtk/maingui.c
@@ -3109,7 +3109,8 @@ mg_create_topwindow (session *sess)
 										  prefs.hex_gui_win_height, 0);
 	sess->gui->window = win;
 	gtk_container_set_border_width (GTK_CONTAINER (win), GUI_BORDER);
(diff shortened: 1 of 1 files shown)


learn-more:
Is this PR still relevant?

PR #2087, opened 2017-11-01
Network list: Add Chatters IRC Network

Shadowgr:
You can see serverlist: http://irc.netsplit.de/servers/?net=Chatters

It's worth mentioning

Chatters is a Greek IRC Network that has been founded since 2010, has experienced staff who has written a lot of patches in bahamut ircd and irc services that can be found at http://bug.chatters.gr

Thanks for your time :)


Arnavion:
Fix it and open a new one.

PR #2104, opened 2017-12-23
Changed AppStream ID to rDNS-style to match what's on Flathub

### LINKED ISSUE #2103: Appstream IDs differ between appdata.xml file and Flatpak metadata
HexChat's Appstream ID on Flathub is `io.github.Hexchat` (https://github.com/flathub/io.github.Hexchat/blob/master/io.github.Hexchat.json#L2)

But HexChat's Appstream file uses a different Appstream ID: `hexchat.desktop` (https://github.com/hexchat/hexchat/blob/master/data/misc/hexchat.appdata.xml.in#L3)

This causes problems. The point of Appstream IDs is for each app has a unique one, but for it to be the same for that app across all distribution channels. If the same app has different Appstream IDs in different distribution channels, then it'll show up as different apps in clients that can display content from multiple sources, such as GNOME Software Center and KDE Discover. Here's what it looks like in KDE discover:

two hexchats

The top one is the Flathub version, the bottom one is provided by my distro's packages.

If you unify the AppStream IDs (ideally standardizing on a reverse-DNS style such as `io.github.Hexchat`), then Discover and GNOME Software (and similar apps) can perform de-duplication correctly, marking both as the same app and offering the user the choice of which version they want to install.

Note: once you change the AppStream ID, you will need to do one of the following to retain the connection between the AppStream data and the desktop file:
- Rename the desktop file to match the AppStream ID (i.e."io.github.Hexchat.desktop")
- Add the following line to hexchat.appdata.xml.in: `<launchable type="desktop-id">hexchat.desktop</launchable>`

Relevant section of the AppStream spec: https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

[note: not an AppStream expert, take this with a grain of salt, etc]

Pointedstick:
Resolves #2103


TingPing:
I think we might as well go all the way and rename the actual desktop file.

Pointedstick:
Your commit doesn't actually fix the problem of mismatched AppStream IDs. You didn't get rid of the trailing `.desktop` part, so the ID still doesn't match Flathub, which uses `io.github.Hexchat`.

HexChat's AppStream ID here should likewise be `io.github.Hexchat`.

TingPing:
The Flathub package does use `io.github.Hexchat.desktop` though?

Pointedstick:
Let me see if we can handle this in Discover.

TingPing:
I'll remove the `.desktop` if appstream recommends it (I pinged hughsie about that in another bug) but I do think this is an issue elsewhere in the stack.

Pointedstick:
We do currently handle this case in Discover, but FWIW, it's not in conformance with the AppStream Spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

>  The `<id>` tag is a unique identifier for this component. It must contain only ASCII characters, dots, hyphens and numbers. Spaces are not allowed.
> The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.Gwenview or com.hugski.ColorHug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.

PR #2113, opened 2018-01-20
Fix building when with-text=true and with-plugin=false

franciscouzo:


TingPing:
The behavior of that arg on `fe-text` is entirely different than `fe-gtk` anyway. So I'd say that this should probably use the same directory: https://github.com/hexchat/hexchat/blob/76430228e12c6b71359a50bf7e8b4c3dd26043c9/src/fe-gtk/fe-gtk.c#L194-L205

PR #2126, opened 2018-02-20
Fix typo in German translation

kbabioch:
The German string contained was prefixed by a `L`, which most probably
was an artefact of copy-pasting the original string. This removes it.


TingPing:
All translations are handled here: https://www.transifex.com/hexchat/hexchat

At a glance this was already fixed there so it will be added to the project next update.

PR #2170, opened 2018-04-07
add gui command "reloadcolors"

sillyslux:
Now you can use `hexchat -e -c "gui reloadcolors"` in a console
to make hexchat reread the colors.conf file after an external change.


src/common/fe.h
```diff
@@ -141,6 +141,7 @@ typedef enum {
 	FE_GUI_FOCUS,
 	FE_GUI_FLASH,
 	FE_GUI_COLOR,
+	FE_GUI_RELOADCOLORS,
```
TingPing: Just add this to the end of the enum to avoid changing the values of everything else.

src/fe-gtk/fe-gtk.c
```diff
@@ -849,6 +849,15 @@ fe_set_throttle (server *serv)
 	}
 }
 
+static void
+fe_reloadcolors (struct session *sess)
+{
+	PrintText (sess, _("Reloading color palette from colors.conf\n"));
+	palette_load();
+	palette_realloc(sess->gui->window);
+	setup_apply_real(0, 0, 0, 0);
```
TingPing: Use `FALSE` not `0`.

sillyslux:
I've tested the patch with master just now and it still works and the two commits are now squashed.
Is there anything else i can do to get this merged soon?

PR #2172, opened 2018-04-07
Make pressing Enter in the serverlist connect.

shlomif:
This is similar to the firefox profile manager and is faster and more
convenient.


shlomif:
Note that I hereby put my changes under the CC0 and/or the MIT / Expat licence and/or the GPL version 2 or above.

TingPing:
I'm not sure that is necessary, *Alt+o* already works and is more discoverable.

shlomif:
@TingPing : hi! sorry for the late reply. Note that I have issues pressing alt+o due to issues with my hands. Also, enter currently does a very little useful thing, so i think enabling it will be more convenient.

shlomif:
Hi all! Can this be merged?

PR #2198, opened 2018-06-01
Add commands list

SoniEx2:
uh, seems to work?

### DIFF
--- a/src/common/plugin.c
+++ b/src/common/plugin.c
@@ -76,6 +76,7 @@ struct _hexchat_list
 	GSList *pos;		/* current pos */
 	GSList *next;		/* next pos */
 	GSList *head;		/* for LIST_USERS only */
+	int ipos;		/* for LIST_COMMANDS_HEXCHAT only */
 	struct notify_per_server *notifyps;	/* notify_per_server * */
 };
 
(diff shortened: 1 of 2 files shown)


src/common/plugin.c
```diff
@@ -1318,11 +1328,41 @@ hexchat_list_free (hexchat_plugin *ph, hexchat_list *xlist)
 int
 hexchat_list_next (hexchat_plugin *ph, hexchat_list *xlist)
 {
-	if (xlist->next == NULL)
-		return 0;
+	int ok = 0; /* this avoids using goto */
+	while (ok^=1) {
```
TingPing: why `^=`

TingPing:
It doesn't seem like a crazy addition but I don't want to add an API for a single plugin and would like to see more use-cases.

src/common/plugin.c
SoniEx2: starts with ok at 0, turns ok into 1 so the loop runs, then when the loop's gonna run the second time it becomes 0 again and continues normally.

this was literally just to avoid using goto, so it uses a loop instead. this is comparable to (but different than) setting it to 1 at the start of the loop, but only uses one line.

SoniEx2:
Well plugins can finally reimplement HELP, add the ability to grep commands' help text, etc.

PR #2206, opened 2018-06-13
fkeys: re-add ctrl+w keybinding

bmintz:
Now that the Ctrl+W key binding is no longer hard coded, users are free to add it in order to close the current tab. However, I believe this should be the default behavior.
After this commit, users can still manually disable the Ctrl+W key binding if they desire.

This is my first commit, so please let me know if I made this change incorrectly.


TingPing:
Shouldn't it be lower case `w`? Anyway I don't have anything against it but it leaves users in a weird spot where some users have the binding and some don't.

PR #2226, opened 2018-07-31
Avoid prioritising MODE queries for channels with hyphens in their name

jrtc27:
If a user has a large number of channels containing hyphens in their names, the initial MODE queries will have the same high priority as any PINGs, and so will block the PINGs from being sent, causing the connection to time out due to a lack of PONGs received.

### DIFF
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -200,13 +200,35 @@ tcp_send_len (server *serv, char *buf, int len)
 	}
 	else
 	{
-		/* WHO/MODE get the lowest priority */
-		if (g_ascii_strncasecmp (dbuf + 1, "WHO ", 4) == 0 ||
-		/* but only MODE queries, not changes */
-			(g_ascii_strncasecmp (dbuf + 1, "MODE", 4) == 0 &&
-			 strchr (dbuf, '-') == NULL &&
-			 strchr (dbuf, '+') == NULL))
+		/* WHO gets the lowest priority */
+		if (g_ascii_strncasecmp (dbuf + 1, "WHO ", 4) == 0)
 			dbuf[0] = 0;
+		/* as do MODE queries (but not changes) */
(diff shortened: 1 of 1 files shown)


jrtc27:
Ping?

src/common/server.c
```diff
@@ -200,13 +200,35 @@ tcp_send_len (server *serv, char *buf, int len)
...
-			 strchr (dbuf, '+') == NULL))
+		/* WHO gets the lowest priority */
+		if (g_ascii_strncasecmp (dbuf + 1, "WHO ", 4) == 0)
 			dbuf[0] = 0;
+		/* as do MODE queries (but not changes) */
+		else if (g_ascii_strncasecmp (dbuf + 1, "MODE", 4) == 0)
+		{
+			char *mode_str, *mode_str_end, *loc;
+			/* skip spaces before channel/nickname */
+			for (mode_str = dbuf + 5; *mode_str == ' '; ++mode_str);
```
TingPing: Is it possible for an empty `"MODE\0"` to make it here?

jrtc27: That's a good question. It's not a valid command (it must always have at least one argument as the user/channel), so we shouldn't be generating it, but I can imagine there's nothing stopping something else from doing it. I suppose we could "cheat" and do `g_ascii_strncasecmp(dbuf + 1, "MODE ", 5)` like with `"WHO "`?

TingPing: Yes that sounds fine.

jrtc27: Forgot to ever get back to this; I have just pushed a fixed version. Having said that, I don't think it would actually have mattered; `MODE\0` would end up with `mode_str` pointing at the `\0`, and `strchr` would then return `NULL`. The main difference is if you ever issue a `MODEFOO` command, which previously (both before this patch and with the previous version of this patch) would have a lower priority. Also we now don't give a `MODE\0` command low priority, which we previously would have done.

PR #2229, opened 2018-08-07
added error for bad reload

A-UNDERSCORE-D:
Added a print for reloading a plugin that doesnt exist. This copies previous behavior and stops confusion

### DIFF
--- a/plugins/python/python.py
+++ b/plugins/python/python.py
@@ -414,7 +414,8 @@ def _on_py_command(word, word_eol, userdata):
         load_filename(filename)
     elif subcmd == 'unload':
         name = ffi.string(word[3]).decode()
-        unload_name(name)
+        if not unload_name(name):
+            print("Can't find a python plugin with that name")
     elif subcmd == 'reload':
         name = ffi.string(word[3]).decode()
         if not reload_name(name):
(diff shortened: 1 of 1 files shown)


plugins/python/python.py
```diff
@@ -417,7 +417,8 @@ def _on_py_command(word, word_eol, userdata):
         unload_name(name)
     elif subcmd == 'reload':
         name = ffi.string(word[3]).decode()
-        reload_name(name)
+        if not reload_name(name):
+            print("cannot find a plugin with that name")
```
TingPing: Lets keep the messages the same as the previous plugin: `Can't find a python plugin with that name`

A-UNDERSCORE-D: changed in f2133bfd011eab013823a568a6e885c21b532458

plugins/python/python.py
```diff
@@ -414,10 +414,12 @@ def _on_py_command(word, word_eol, userdata):
...
     elif subcmd == 'unload':
         name = ffi.string(word[3]).decode()
-        unload_name(name)
+        if not unload_name(name):
+            print("Can't find a python plugin with that name")
     elif subcmd == 'reload':
         name = ffi.string(word[3]).decode()
-        reload_name(name)
+        if not reload_name(name):
+            print("Can't find a python plugin with that name")
```
TingPing: Single quotes, also use hexchat_print as shown below.

PR #2230, opened 2018-08-08
fixed /exec behaviour (again)

A-UNDERSCORE-D:
resolves issues where using 'multiline' statements cause syntax errors:

```python
 From cffi callback <function _on_py_command at 0x7f29afe40378>:
 Traceback (most recent call last):
   File "<init code for '_hexchat_embedded'>", line 411, in _on_py_command
   File "<init code for '_hexchat_embedded'>", line 369, in exec_in_interp
   File "<init code for '_hexchat_embedded'>", line 100, in compile_line
   File "<string>", line 1
     for x in "test": print(x)
                             ^
 SyntaxError: unexpected EOF while parsing
```

### DIFF
--- a/plugins/python/python.py
+++ b/plugins/python/python.py
@@ -97,7 +97,8 @@ else:
         return compile(data, filename, 'exec', optimize=2, dont_inherit=True)
 
     def compile_line(string):
-        return compile(string, '<string>', 'single', optimize=2, dont_inherit=True)
+        # using exec over single as single causes weird behaviour
+        return compile(string, '<string>', 'exec', optimize=2, dont_inherit=True)
 
 
 class Plugin:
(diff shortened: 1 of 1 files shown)


plugins/python/python.py
```diff
@@ -97,8 +97,8 @@ def compile_file(data, filename):
         return compile(data, filename, 'exec', optimize=2, dont_inherit=True)
 
     def compile_line(string):
-        # using exec over single as single causes weird behaviour
-        return compile(string, '<string>', 'exec', optimize=2, dont_inherit=True)
+        # newline appended to solve unexpected EOF issues
+        return compile(string + "\n", '<string>', 'single', optimize=2, dont_inherit=True)
```
TingPing: Single quotes!

A-UNDERSCORE-D: fixed in 88d259e7

PR #2238, opened 2018-08-18
Various fixes (errors in hexchat-text on windows, and crlf issues in server list)

micronn:
The call to g_io_channel_set_buffered with encoding not being NULL was failing and producing lots of error messages in hexchat-text. Btw, I'm not sure if we really need unbuffered i/o here and reads byte by byte. But, well... it works.

The other commit just skips \r when reading the server list file (and also the startup.txt). Other files are not created with fwrite and shouldn't have this issue unless the user edits them manually.

### DIFF
--- a/src/common/outbound.c
+++ b/src/common/outbound.c
@@ -2560,7 +2560,6 @@ gboolean
 load_perform_file (session *sess, char *file)
 {
 	char tbuf[1024 + 4];
-	char *nl;
 	FILE *fp;
 
 	fp = hexchat_fopen_file (file, "r", 0);		/* load files from config dir */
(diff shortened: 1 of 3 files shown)


src/common/outbound.c
```diff
@@ -2570,11 +2569,9 @@ load_perform_file (session *sess, char *file)
 	tbuf[1024] = 0;
 	while (fgets (tbuf, 1024, fp))
 	{
-		nl = strchr (tbuf, '\n');
-		if (nl == tbuf) /* skip empty commands */
+		tbuf[strcspn (tbuf, "\r\n")] = 0;
```
TingPing: Well that just breaks Unix.

I think you should just ensure files on Windows are written using `\n` only.

TingPing: Ah my mistake. I do still think there is an advantage to simply using `\n` everywhere.

PR #2255, opened 2018-10-06
Add a channel context menu item to show channels

pabs3:


TingPing:
I wonder if this logic should not use `doat` and just do the work itself but maybe this is just more simple.

This should also probably try to be smart about only showing this option when it knows you are in the channel. (Seems the cycle option does this).

TingPing:
Oh and the CI failure is relevant, seems you never tested this.

pabs3:
Could you indicate where in the code the Cycle Channel menu item checks
for presence in the channel? I wasn't able to find that.

Re doat, I could make it call fe_ctrl_gui instead but then I'd have to
re-implement parts of the cmd_doat function so I'd prefer to keep it
as-is. I cannot call cmd_doat directly because it is a static function
so only works when called from outbound.c.

I indeed hadn't tested it, fixed the issue now.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

pabs3:
I've tested the current patch.

As I mentioned before, I prefer to use doat than to re-implement parts
of the cmd_doat function, I hope you understand.

It does not show the Show Channel item when you have not joined the
channel, as you requested.

It doesn't hide the Show Channel item when you choose bring up a
context menu for text of the current channel, do you think I need to
hide the item in that case?

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

src/fe-gtk/menu.c
```diff
@@ -1035,6 +1047,8 @@ menu_chanmenu (struct session *sess, GdkEventButton * event, char *chan)
 												 str_copy);
 	else
 	{
+		menu_quick_item_with_callback (menu_chan_show, _("Show Channel"), menu,
```
TingPing:
```suggestion
		menu_quick_item_with_callback (menu_chan_show, _("Focus Channel"), menu,
```

pabs3:
Updated the language to use Focus instead of Show.

TingPing:
> It doesn't hide the Show Channel item when you choose bring up a
context menu for text of the current channel, do you think I need to
hide the item in that case?

Oh I probably should have re-read the thread. This would indeed be nice if you are still interested.

pabs3:
Is there a function that can determine if the current channel matches a channel name text? Also, it seems like current_sess->channel is the way to detect the current channel?

TingPing:
The `find_channel()` call a few lines up returns a session object, compare that against `current_sess`.

Issue #1393, opened 2015-06-11
Make IRCv3 message tags accessible to scripts


RenaKunisaki:
From what I can see, the Python interface offers no way to access the IRCv3 tags of a received message.

Example message:
```
@color=#FF4500;display-name=PumpkinAppearifier;emotes=;subscriber=0;turbo=0;user-type=;user_type= :pumpkinappearifier!pumpkinappearifier@pumpkinappearifier.tmi.twitch.tv PRIVMSG #themartonfi1228 :Hey Marton
```

Even the `RAW LINE` server hook doesn't include these in `word`, `word_eol` or (as best I can tell) `attributes` parameters (though `attributes` appears not to be iterable so it's hard to say for sure).

MikeZeDev:
I second that. This would be useful , at least for Twitch. I am using lua.

TingPing:
Would basically require a new API. The attributes api we have is not flexible enough for this.

In another project of mine I have a parser that makes a hashtable of the tags though: https://github.com/TingPing/irc-client/blob/master/src/irc-message.c#L116-L176

gpnoel:
Will this become a feature in the future, not necessarily soon but just at some point? Also, is that parser of yours able to be used in conjunction with HexChat?

TingPing:
No active work has started on it but it could happen.

PR #2256, opened 2018-10-06
Allow Python interface to access IRCv3 information

Yaulendil:
A small change to forward IRCv3 information received by Hexchat to Python scripts that request it. This addresses, though perhaps not fully, #1393.

In Python, the function passed to `hexchat.hook_server_attrs()` now receives the IRCv3 message as a string under the Attribute object, as `Attribute.ircv3`. This should not break any Python addons, unless they call `dir(Attribute)` and expect a very specific return value (I do not know of any that do).

It does not affect any of the other scripting interfaces because I dont know those interfaces as well.
It does not allow Python to access the IRCv3 information through `hexchat.hook_print_attrs()` because I wanted, at least at first, to keep my edits to the code minimal.

I have been running this fork for a week now, with an addon that heavily utilizes the change, and it seems to be quite stable.

### DIFF
--- a/plugins/python/python.c
+++ b/plugins/python/python.c
@@ -998,7 +998,6 @@ Attribute_New(hexchat_event_attrs *attrs)
 	attr = PyObject_New(AttributeObject, &Attribute_Type);
 	if (attr != NULL) {
 		attr->time = PyLong_FromLong((long)attrs->server_time_utc);
-		char* *tagstring = attrs->ircv3_line;
 		attr->ircv3 = PyUnicode_FromString(attrs->ircv3_line);
 	}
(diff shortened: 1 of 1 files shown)


TingPing:
I strongly dislike just shoving the line in there and calling it a day. If we want to expose raw tags it should be in a hash table or so.

SoniEx2:
It's a good idea to make the raw string available as well, in case anyone relies on the wire-order for some reason.

Davarice:
While I do see the value of the raw string being available, I do agree that it would probably be more performant to be done as a hash table in C, so that plugins dont need to spend time splitting the string up themselves.

PR #2259, opened 2018-10-13
Fix UB (dangling pointer usage)

SoniEx2:
Using dangling pointers, even if it's just to compare them with valid
pointers (such as comparing them with the ones in sess_list), is UB.

This fixes that.

Oh yeah and btw this is completely untested, but it didn't have any compile warnings or errors so I assume it's good, right?

### DIFF
--- a/src/common/plugin.c
+++ b/src/common/plugin.c
@@ -654,6 +654,18 @@ plugin_emit_print (session *sess, char *word[], time_t server_time)
 							HOOK_PRINT | HOOK_PRINT_ATTRS);
 }
 
+/* used by plugin_emit_dummy_print to fix some UB */
+static void
+check_and_invalidate(void *plug_, void *killsess_)
+{
+	hexchat_plugin *plug = plug_;
+	session *killsess = killsess_;
+	if (plug->context == killsess)
+	{
+		plug->context = NULL;
+	}
+}
+
 int
 plugin_emit_dummy_print (session *sess, char *name)
 {
@@ -664,7 +676,16 @@ plugin_emit_dummy_print (session *sess, char *name)
 	for (i = 1; i < 32; i++)
 		word[i] = "\000";
 
-	return plugin_hook_run (sess, name, word, NULL, NULL, HOOK_PRINT);
+	i = plugin_hook_run (sess, name, word, NULL, NULL, HOOK_PRINT);
+
+	/* shoehorned fix for Undefined Behaviour */
+	/* see https://stackoverflow.com/q/52628773/3691554 */
+	/* this needs to be done on the hexchat side */
+	if (strcmp(name, "Close Context") == 0) {
+		g_slist_foreach(plugin_list, &check_and_invalidate, sess);
+	}
+
+	return i;
 }
 
 int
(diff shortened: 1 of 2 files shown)


src/common/plugin.c
```diff
@@ -664,7 +676,16 @@ plugin_emit_dummy_print (session *sess, char *name)
 	for (i = 1; i < 32; i++)
 		word[i] = "\000";
 
-	return plugin_hook_run (sess, name, word, NULL, NULL, HOOK_PRINT);
+	i = plugin_hook_run (sess, name, word, NULL, NULL, HOOK_PRINT);
+
+	/* shoehorned fix for Undefined Behaviour */
+	/* see https://stackoverflow.com/q/52628773/3691554 */
```
TingPing: This post is full of nonsense and confusion. What are you actually fixing?

SoniEx2: So, here's what the general plugin workflow looks like. You call a plugin, it gets a context, etc.

The plugin can call get_context to retrieve the context. This can then be stored somewhere. If the context is closed, this becomes dangling - but that's okay, we can use the "Close Context" hook to clean them up so they don't become dangling!

However, when the "Close Context" hook returns, plugins (as in, `hexchat_plugin` objects) may still have the dangling context inside them. As such, the plugin's next call to get_context could trigger UB.

By setting the plugin's context to NULL, the next call would return NULL, which *is* a valid pointer - although *dereferencing* it is still UB, it's completely safe to store it, copy it around and compare it to other valid pointers!

TingPing: The better solution is just ensure all calls are protected by `is_session()` better.

SoniEx2: You still need to clear the dangling pointers, as even passing them to `is_session()` is UB.

`is_session(dangling_pointer)` is UB.

TingPing: No it isn't.

SoniEx2: The stackoverflow, quoting the C standard, says it is. So, yes, it is.

TingPing: That makes no sense at all, the value of a pointer is just an address, it never changes, why would that ever make sense for any C implementation?

Anyway it works fine and has worked fine for probably 40 years on every compiler we ever have and ever will support.

SoniEx2: here:

https://internals.rust-lang.org/t/comparing-dangling-pointers/3019

https://internals.rust-lang.org/t/pointers-are-complicated-or-whats-in-a-byte/8045

now *read*.

TingPing: Ok there was one reply in that entire post that made sense. `malloc()` could technically return the same address. That is the most unlikely of problems but sounds possible. Anything else just sounds like nonsense.

SoniEx2: "malloc() could technically return the same address" just like returning a pointer to inside the current stack frame and passing that pointer to another function. this is literally what UB looks like.

SoniEx2: "Any use of a pointer that refers to freed space results in undefined behavior." - free(3p) - POSIX Programmer's Manual, 2013

Issue #2023, opened 2017-07-05
A change of font & its size not well applied to server list and user list


BBasile:
Under Windows 7, there's something wrong with the height of the items, let's say if you go from 10 to 15, the user list and the server list items height is not updated but the text has the good size.

learn-more:
I would assume that the item height is calculated once, and not properly re-calculated when font size changes.

TingPing:
Just a guess but maybe its caused because we set it as fixed height: https://github.com/hexchat/hexchat/blob/15600f405f2d5bda6ccf0dd73957395716e0d4d3/src/fe-gtk/chanview-tree.c#L156

TingPing:
@BBasile This issue is not solved, please stop closing it.

Martchus:
Yes, it isn't solved. And the issue is also not limited to Windows. Since the issue might be good for a first commit, so I'm going to look into it :-)

PR #2264, opened 2018-10-31
Apply font changes to server list

Martchus:
* Prevent use of inflexible
  gtk_cell_renderer_text_set_fixed_height_from_font()
  (performance with lots of servers still seems good)
* Redraw server list when applying setup

See https://github.com/hexchat/hexchat/issues/2023

---

I'm aware this fixes only one of the lists in the issue. But I would like to wait for feedback first because it is my first time working GTK.

### DIFF
--- a/src/fe-gtk/chanview-tree.c
+++ b/src/fe-gtk/chanview-tree.c
@@ -153,7 +153,6 @@ cv_tree_init (chanview *cv)
 	renderer = gtk_cell_renderer_text_new ();
 	if (prefs.hex_gui_compact)
 		g_object_set (G_OBJECT (renderer), "ypad", 0, NULL);
-	gtk_cell_renderer_text_set_fixed_height_from_font (GTK_CELL_RENDERER_TEXT (renderer), 1);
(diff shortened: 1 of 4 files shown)


src/fe-gtk/chanview-tree.c
```diff
@@ -394,3 +393,14 @@ cv_tree_is_collapsed (chan *ch)
 	
 	return ret;
 }
+
+static void
+cv_tree_queue_draw (chanview *cv)
+{
+	/* assign the model again to cause the tree view to be redrawn
+	   note: Simply using gtk_widget_queue_draw() is not sufficient. */
```
TingPing: This seems like a hacky workaround, I'd prefer figuring out a way to just tell it to redraw.

Martchus: I've looked at the documentation and tried several things before coming up with that hack. Do you have a specific API function in mind? I only found that function I mention in the comment but it is not sufficient because the layout is still not updated.

The preferred way for this to work is likely that GTK triggers the layout update itself if relevant properties like the font change. So doing things manually here will always be not the optimal solution and therefore a "hacky workaround".

PR #2272, opened 2018-11-22
servlist: add hackint irc network

mweinelt:
- requires the use of TLS to connect on port 6697
- supports and encourages authentication via SASL PLAIN and EXTERNAL

Certificates used are signed by Let's Encrypt.

### DIFF
--- a/src/common/servlist.c
+++ b/src/common/servlist.c
@@ -193,6 +193,11 @@ static const struct defaultserver def[] =
 #endif
 	{0,			"irc.globalgamers.net"},
 
+#ifdef USE_OPENSSL
+	{"hackint", 0, 0, 0, LOGIN_SASL, 0, TRUE},
+	{0,			"irc.hackint.org"},
+#endif
+
 	{"Hashmark",	0},
 	{0,			"irc.hashmark.net"},
 


mweinelt:
@TingPing I would appreciate some feedback to get this merged.

src/common/servlist.c
```diff
@@ -196,6 +196,12 @@ static const struct defaultserver def[] =
 #endif
 	{0,			"irc.globalgamers.net"},
 
+#ifdef USE_OPENSSL
+	{"hackint", 0, 0, 0, LOGIN_SASL, 0, TRUE},
+	{0,			"irc.hackint.org/+6697"},
```
TingPing: Drop the explicit port (is the default for TLS), and do we really need two different hostnames?

mweinelt: Rebased the pr and dropped the secondary hostname and port.

src/common/servlist.c
```diff
@@ -193,6 +193,11 @@ static const struct defaultserver def[] =
 #endif
 	{0,			"irc.globalgamers.net"},
 
+#ifdef USE_OPENSSL
```
TingPing: Unless it supports *only* TLS you can drop the ifdef too

mweinelt: Yes, we are TLS only.

mweinelt:
Thanks!

PR #2274, opened 2018-11-23
Update Translations : Simple Chinese

yichangshengwu:
some menu item was still in English, got some of them pathed.


TingPing:
All translations go through Transifex: https://www.transifex.com/hexchat/hexchat

yichangshengwu:
some words appeared times but only one translated in the “.po” file !!!
so that, the same words appeared in dfferent languages at different menu .


then, although the words were translated one time in the ".po" file, but the menu looks like this:

PR #2278, opened 2018-11-30
Bump minimum Meson version to 0.46

rzumer:
Solves the following warning when calling `meson`:
```
WARNING: Project specifies a minimum meson_version '>= 0.40.0' but uses features which were added in newer versions:
 * 0.46.0: {'compiler.has_link_argument', 'compiler.has_multi_link_argument'}
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('hexchat', 'c',
   version: '2.14.2',
-  meson_version: '>= 0.40.0',
+  meson_version: '>= 0.46.0',
   default_options: [
     'c_std=gnu89',
     'buildtype=debugoptimized',


rzumer:
I see that it is more complicated than expected due to the Travis configuration, my bad. I am not sure why Meson complains even though it compiles with 0.40, but I do not want to start upgrading the Travis configuration if it is not wanted, so will close until sorted out.

TingPing:
Meson doesn't understand those calls are guarded behind version checks and upstream should better handle that.

rzumer:
See https://github.com/mesonbuild/meson/issues/3824

PR #2279, opened 2018-11-30
Ignore update/upgrade build directories

rzumer:
The standard build process generates these directories, so add them to `.gitignore`.


TingPing:
What command generates those?

rzumer:
@TingPing Apologies, these were Meson-generated directories, which should not appear using the documented build process. I should have double-checked.

PR #2289, opened 2018-12-26
made sure to set sys.argv if it is not set. fixes #2282

### LINKED ISSUE #2282: Python sys.argv is uninitialized when using cffi
It should be initialized to `[""]` to prevent errors in some libraries

A-UNDERSCORE-D:
Name set to match [old behavior](https://github.com/hexchat/hexchat/blob/02c92599faa8adbe8710177f943c3ca3cb4a65d9/plugins/python/python.c#L2693)

### DIFF
--- a/plugins/python/python.py
+++ b/plugins/python/python.py
@@ -10,6 +10,9 @@ import traceback
 import weakref
 from _hexchat_embedded import ffi, lib
 
+if not hasattr(sys, 'argv'):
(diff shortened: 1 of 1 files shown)


TingPing:
Can you please prefix commits with `python: ` to fit in with existing style.

A-UNDERSCORE-D:
Done

PR #2290, opened 2018-12-26
Make sure `help()` doesn't cause hexchat to hang

linuxdaemon:
Replace `pydoc.help` with a copy of `pydoc.Helper` with an empty
`StringIO` instead of stdin


TingPing:
Just to confirm, tested on py2 and py3?

linuxdaemon:
Now it is

PR #2291, opened 2018-12-26
Make the python plugins table dynamically sized

linuxdaemon:
Adjust the width of the columns depending on the length of the data in
each element.

### DIFF
--- a/plugins/python/python.py
+++ b/plugins/python/python.py
@@ -349,15 +349,33 @@ def list_plugins():
         lib.hexchat_print(lib.ph, b'No python modules loaded')
         return
 
-    lib.hexchat_print(lib.ph, b'Name         Version  Filename             Description')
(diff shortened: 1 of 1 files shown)


plugins/python/python.py
```diff
@@ -346,15 +346,33 @@ def list_plugins():
...
-        lib.hexchat_print(lib.ph, string)
+        tbl.append((name, version, basename, description))
+
+    column_sizes = [
+        max(len(item) for item in column)
+        for column in zip(*tbl)
+    ]
+
+    for row in tbl:
+        lib.hexchat_print(
```
TingPing: Just to nitpick I really dislike this indentation. Just do:

```python
     lib.hexchat_print(lib.ph, b' '.join(item.ljust(column_sizes[i])
                                         for i, item in enumerate(row)))
```

That turns 7 lines into 2...

PR #2292, opened 2018-12-26
Python plugin cleanup

A-UNDERSCORE-D:
Fixes some readability, removes some inefficient code

### DIFF
--- a/plugins/python/_hexchat.py
+++ b/plugins/python/_hexchat.py
@@ -1,6 +1,7 @@
-from contextlib import contextmanager
 import inspect
 import sys
+from contextlib import contextmanager
+
 from _hexchat_embedded import ffi, lib
 
 __all__ = [
@@ -40,13 +41,15 @@ def __get_current_plugin():
     while '__plugin' not in frame.f_globals:
         frame = frame.f_back
         assert frame is not None
+
     return frame.f_globals['__plugin']
 
 
 # Keeping API compat
-if sys.version_info[0] is 2:
+if sys.version_info[0] == 2:
     def __decode(string):
         return string
+
 else:
     def __decode(string):
         return string.decode()
@@ -64,16 +67,18 @@ def emit_print(event_name, *args, **kwargs):
         arg = args[i].encode() if len(args) > i else b''
         cstring = ffi.new('char[]', arg)
         cargs.append(cstring)
-    if time is 0:
+
+    if time == 0:
         return lib.hexchat_emit_print(lib.ph, event_name.encode(), *cargs)
-    else:
-        attrs = lib.hexchat_event_attrs_create(lib.ph)
-        attrs.server_time_utc = time
-        ret = lib.hexchat_emit_print(lib.ph, attrs, event_name.encode(), *cargs)
-        lib.hexchat_event_attrs_free(lib.ph, attrs)
-        return ret
+
+    attrs = lib.hexchat_event_attrs_create(lib.ph)
+    attrs.server_time_utc = time
+    ret = lib.hexchat_emit_print(lib.ph, attrs, event_name.encode(), *cargs)
+    lib.hexchat_event_attrs_free(lib.ph, attrs)
+    return ret
 
 
+# TODO: this shadows itself. command should be changed to cmd
 def command(command):
     lib.hexchat_command(lib.ph, command.encode())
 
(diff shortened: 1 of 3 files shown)


plugins/python/_hexchat.py
```diff
@@ -64,16 +66,18 @@ def emit_print(event_name, *args, **kwargs):
         arg = args[i].encode() if len(args) > i else b''
         cstring = ffi.new('char[]', arg)
         cargs.append(cstring)
-    if time is 0:
+
+    if time == 0:
```
TingPing: Not sure the value of changing all the `is` usage, `==` is just slower.

plugins/python/python.py
```diff
@@ -346,15 +380,35 @@ def list_plugins():
         lib.hexchat_print(lib.ph, b'No python modules loaded')
         return
 
-    lib.hexchat_print(lib.ph, b'Name         Version  Filename             Description')
-    lib.hexchat_print(lib.ph, b'----         -------  --------             -----------')
+    lib.hexchat_print(
```
TingPing: I think this mostly hinders readability.

plugins/python/_hexchat.py
A-UNDERSCORE-D: `is` is an identity check, so it only works between [-5 and 256](https://docs.python.org/3/c-api/long.html#c.PyLong_FromLong) as those are cached in the small integer cache. if a future change removes or changes the this could break. the `is` check pretty much does `id(a) == id(b)`, so its not really any slower afaik

plugins/python/python.py
A-UNDERSCORE-D: Yeah, Im not 100% sure I like that, it is required to keep it within PEP8 line length limits

plugins/python/_hexchat.py
TingPing: The world would explode if it was changed so it will never change.

plugins/python/python.py
A-UNDERSCORE-D: changed in later commit

plugins/python/_hexchat.py
A-UNDERSCORE-D: This is true. Not sure if it helps readability, though what it WILL do is cause inconsistencies if we ever do that check on something bigger than 256

plugins/python/_hexchat.py
```diff
@@ -97,21 +102,24 @@ def get_info(name):
...
+    if _type == 1:
         return __decode(ffi.string(string_out[0]))
-    elif type is 2 or type is 3:  # XXX: 3 should be a bool, but keeps API
+
+    if _type in (2, 3):  # XXX: 3 should be a bool, but keeps API
         return int_out[0]
-    else:
-        assert False
+
+    raise ValueError('Unknown storage type for pluginpref')
```
TingPing: Revert this. If any other value exists its a programmer error on hexchat's side, thus an assertion is appropriate.

plugins/python/python.py
```diff
@@ -285,40 +299,46 @@ def _on_say_command(word, word_eol, userdata):
...
+
     return 0
 
 
 def load_filename(filename):
     filename = os.path.expanduser(filename)
     if not os.path.isabs(filename):
-        configdir = __decode(ffi.string(lib.hexchat_get_info(lib.ph, b'configdir')))
+        configdir = __decode(
+            ffi.string(lib.hexchat_get_info(lib.ph, b'configdir'))
```
TingPing: left-over style change.

plugins/python/python.py
```diff
@@ -428,23 +452,34 @@ def _on_py_command(word, word_eol, userdata):
...
     elif subcmd == 'console':
         lib.hexchat_command(lib.ph, b'QUERY >>python<<')
+
     elif subcmd == 'list':
         list_plugins()
+
     elif subcmd == 'about':
-        lib.hexchat_print(lib.ph, b'HexChat Python interface version ' + VERSION)
+        lib.hexchat_print(
+            lib.ph, b'HexChat Python interface version ' + VERSION
```
TingPing: left-over style change.

plugins/python/_hexchat.py
TingPing: Alright

plugins/python/_hexchat.py
A-UNDERSCORE-D: In which case a message with the assert should be warranted. What about `raise AssertionError("Unknown type stored for pluginpref")`? that way its more than just an assertion error with a stack trace

TingPing: Thats fine.

A-UNDERSCORE-D: Switched to `raise AssertionError`

PR #2294, opened 2019-01-05
Remove colon from trailing parameters

### LINKED ISSUE #2271: Various command parsers do not handle a colon before the last parameter
This issue was first noticed when testing Hexchat with Inspircd v3, channel topic times keep showing as `0` in unix time. `CHGHOST` also does not seem to remove the colon before the last parameter resulting in a user's host being set to `@:<newhost>`.
Inspircd 3.0 changed its message formats a bit, resulting in a colon being prepended to the last parameter in every message from the server. According to the IRC RFCs, clients should generally just ignore the colon before the last parameter in all cases.
## Versions
HexChat: 2.12.4 and 2.14.1
[InspIRCd](https://github.com/inspircd/inspircd) built from commit https://github.com/inspircd/inspircd/tree/37977710a79472741c043b2dc8c17f520429c9f4

linuxdaemon:
Make sure trailing parameters are treated the same as other parameters
to avoid issues like #2271

Fixes #2271 

This is not the cleanest fix in the world, but based on my (albeit limited) testing it fixes the issue and doesn't seem to introduce more.

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -1557,6 +1557,27 @@ irc_inline (server *serv, char *buf, int len)
 
 	/* split line into words and words_to_end_of_line */
 	process_data_init (pdibuf, buf, word, word_eol, FALSE, FALSE);
+	// Ignore the prefix
+	char **trail = word_eol + 1;
+
+	// Iterate until the next element is NULL or {0, }
+	while (*trail && (*trail)[0] && (*trail)[0] != ':')
+		trail++;
+
+	if (*trail && (*trail)[0] == ':')
+		// Move the trail forward one character
+		(*trail)++;
+
+	// Ignore the prefix
+	trail = word + 1;
+
+	// Iterate until the next element is NULL or {0, }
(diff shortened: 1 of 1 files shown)


TingPing:
I find this change pretty invasive and scary. Currently hexchats text handling effectively has no concept of 'trailing' so by modifying the text globally it could have who knows what side-effects since nothing is properly tested.

linuxdaemon:
The handlers shouldn't differentiate parameters from trailing, this just removes the `:` before the last parameter if it's there which as long as none of the handlers explicitly look for it, should be fine

TingPing:
> The handlers shouldn't differentiate parameters from trailing, this just removes the : before the last parameter if it's there which as long as none of the handlers explicitly look for it, should be fine

All it takes is a single function doing: `foo = word[1] == ':' ? word[1] : word_eol[1]` for it to be wrong. And I'm sure neither of us have audited every path to ensure that.

Also since you are actually modifying the raw text this happens before plugins get it so you are basically breaking API in my eyes.

Issue #2296, opened 2019-01-16
HexChat needs restarting when docking/undocking with hi-res display


dwmw2:
My laptop has a hi-resolution display which is scaled to 200% when I'm using it on its own.
When docked, the only sane configuration I've found is 100% scaling, and to change the laptop display to "low-resolution" 1920x1200 mode.

HexChat does not cope with the scaling changing.

If started while docked (at 100% i.e. no scaling), then undocked, each line in the chat window is about half as high as it should be. I can only see the bottom half of each character.
If started while undocked (at 200%) then docked, each line becomes twice the height it should have been; as if the lines in the text are double-spaced.

TingPing:
This is on Windows?

dwmw2:
No, Linux (Ubuntu 18.04). Sorry for not specifying.

TingPing:
Well I am going to close this in favor of #828.

HexChat isn't DPI aware at all and does no scaling itself. Whatever scaling you see is happening somewhere else not to its knowledge.

dwmw2:
#828 really does seem to be Windows-specific; I'm not sure if it's the same thing.

The main problem is the text in the main text box, and it seems I can work around that by changing the font. The font size information (in pixels) is cached in backend_font_open() and once that is called again, text spacing goes back to normal.

The text in the userlist does actually double in size when I move to a non-scaled screen (as opposed to just being double-spaced but still the same size font). Changing the font makes it end up normal sized but double-spaced — i.e. still in the same place but smaller; there's something else going on there.

dwmw2:
For the userlist, I suspect if I could make it redo the `gtk_cell_renderer_text_set_fixed_height_from_font()` calls in `userlist_add_columns()`, it might behave?

TingPing:
That issue discussion is very Windows specific but the concept should apply everywhere, HexChat doesn't do scaling of any rendering.

Now I guess it is possible some X11 properties are changing automatically that get picked up by GTK though.

dwmw2:
HexChat is caching the height of the configured font in pixels.

When that changes due to DPI changing, GTK does all its rendering right but HexChat is still using the old cached pixel height values, so it places things in the wrong place.

PR #2299, opened 2019-01-21
Recalculate font metrics when DPI changes

dwmw2:
Fixes: #2296


EwoutH:
Can we merge this? I don't like this look :)

screenshot_156

PR #2301, opened 2019-01-29
Remove : from various trailing parameters

linuxdaemon:
Partial fix for #2271 

This isn't an exhaustive list, but it's everything I could find. The bug still exists in the parser though, this is just a workaround for the moment

### DIFF
--- a/src/common/modes.c
+++ b/src/common/modes.c
@@ -735,6 +735,8 @@ handle_mode (server * serv, char *word[], char *word_eol[],
 		if (!(*word[i + offset]))
 			break;
 		num_args++;
+		if (word[i + offset][0] == ':')
+			break;
 	}
 
 	/* count the number of modes (without the -/+ chars */
--- a/src/common/modes.c
+++ b/src/common/modes.c
@@ -767,7 +767,7 @@ handle_mode (server * serv, char *word[], char *word_eol[],
 				arg++;
 				// is this the last argument?
 				if (arg == num_args + 1)
-					argstr = STRIP_COLON(word[arg + offset]);
+					argstr = STRIP_COLON(word, word_eol, arg+offset);
 				else
 					argstr = word[arg + offset];
 			}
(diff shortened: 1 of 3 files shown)


TingPing:
This usage is wrong throughout, when `:` is present you would use `word_eol` not `word`.

linuxdaemon:
Ah woops, I had copied the logic from the 329 case

src/common/modes.c
```diff
@@ -765,7 +765,11 @@ handle_mode (server * serv, char *word[], char *word_eol[],
 			if ((all_modes_have_args || mode_has_arg (serv, sign, *modes)) && arg < (num_args + 1))
 			{
 				arg++;
-				argstr = word[arg + offset];
+				// is this the last argument?
```
TingPing: Could you elaborate on this. This seems to be a behavior change beyond the trailing `:`

linuxdaemon: This is to remove the trailing colon in a message like `MODE #channel +v user user :user`

TingPing: That seems wrong.

The first arg that starts with `:` becomes the trailing arg.

linuxdaemon: I think combining it with a check in the arg_count loop for `:` should work, that would ensure the last argument is always the trailing argument and would handle it all as one

TingPing: Sure I think it will work but it is a violation of the protocol and since this whole discussion is about that I don't think it should be done.

linuxdaemon: in what way? the check in the loop would make it so the argument with the `:` would be considered the last arg and it would use word_eol for that. Having a trailing parameter on `MODE` is valid

linuxdaemon: InspIRCd v3 sends `MODE #channel +oo user :user1`, each mode parameter being a new command parameter, and the last one being a trailing param. The check in the loop ensures the arg_count reflects the actual number (ie 2 for `MODE +gg word :multi word`)

TingPing: > The check in the loop ensures the arg_count reflects the actual number (ie 2 for MODE +gg word :multi word)

Yes but that ensuring is incorrect. `MODE +gggg :a b c d` According the protocol that is `a b c d` not 4 args.

TingPing: hmm, I see you fixed `num_args` above. So assuming that is all correct why does the `if (arg == num_args + 1)` check need to happen? It would be safe to call `STRIP_COLON` each arg.

PR #2309, opened 2019-02-25
Extend Adwaita workaround [#2305]

### LINKED ISSUE #2305: Ubuntu 18.10 theme mismatch, input and title boxes white
How it looks in 18.04:
image

How it looks in 18.10:
image

Note the title/user bar at the top and the input box at the bottom are both white now.

I recompiled with source just to get to the theme manager, which made no difference. Open to ideas, but GUI work is completely out of my wheelhouse. #1832 offers a temporary solution that makes the input box text legible, but I'm assuming another similar fix needs to be made for the new version of GNOME?

cajuncooks:
Resolves #2305.

### DIFF
--- a/src/fe-gtk/fe-gtk.c
+++ b/src/fe-gtk/fe-gtk.c
@@ -293,10 +293,10 @@ create_input_style (GtkStyle *style)
 		GtkSettings *settings = gtk_settings_get_default ();
 		char *theme_name;
 
-		/* gnome-themes-standard 3.20 relies on images to do theming
+		/* gnome-themes-standard 3.20+ relies on images to do theming
 		 * so we have to override that. */
 		g_object_get (settings, "gtk-theme-name", &theme_name, NULL);
-		if (g_str_has_prefix (theme_name, "Adwaita"))
+		if (g_str_has_prefix (theme_name, "Adwaita") || g_str_has_prefix (theme_name, "Yaru"))
 			gtk_rc_parse_string (adwaita_workaround_rc);
 		g_free (theme_name);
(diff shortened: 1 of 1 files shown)


src/fe-gtk/fe-gtk.c
```diff
@@ -293,10 +293,10 @@ create_input_style (GtkStyle *style)
 		GtkSettings *settings = gtk_settings_get_default ();
 		char *theme_name;
 
-		/* gnome-themes-standard 3.20 relies on images to do theming
+		/* gnome-themes-standard 3.20+ relies on images to do theming
 		 * so we have to override that. */
 		g_object_get (settings, "gtk-theme-name", &theme_name, NULL);
-		if (g_str_has_prefix (theme_name, "Adwaita"))
```
TingPing: Can you test Yaru-dark also and change it to a prefix check like Adwaita?

TingPing:
Also nitpick but I'd prefer the commit message:

```
Extend input box GTK theme workaround to include Yaru

Fixes #2305
```

src/fe-gtk/fe-gtk.c
cajuncooks: I ran `gsettings set org.gnome.desktop.interface gtk-theme Yaru-dark` and confirmed application response (Chromium turned dark, text boxes in question did too, but lost their theme). Changing to a prefix check resolves this, and maintains the theme through a gtk-theme update while Hexchat is open. I've also updated the commit message.

TingPing:
Thanks!

PR #2314, opened 2019-03-17
Update ru.po

mandala-mbwa:
fix translation


TingPing:
All translations are handled here: https://www.transifex.com/hexchat/hexchat/

PR #2316, opened 2019-03-27
Avoid using ELLIPSIS macro

### LINKED ISSUE #2275: i think there is a bug somewhere...
if,
no matter what language the user's DE is, those menu item below always displays in English,
{
    He_xchat -> _Load Plugin or Script...
    _Server -> _Join a Channel...
    _Window -> Save Text...
    ...
}
then,
there must be a bug somewhere!
- - - 
there was five items like this, the same thing of them is,
in the ./src/fe-gtk/menu.c file, there was a "ELLIPSIS",
i don't know what caused this exactly,
so the only thing i can do is write them down here...

memnoth:
Some strings using the macro cannot be translated in the right way,
because PO files are made by source code instead of compile-time code.
The macro is actually translated while compilation and therefore it
causes the issue, #2275.

To fix the issue, use the literal ellipsis string instead of it.

Fixes #2275


TingPing:
The reason this was done in the first place is msvc doesn't like non-ascii text.

memnoth:
I changed it to use ellipsis utf-8 hex code itself in the messages. This is probably a solution for windows too. Another solution is to use three dots there.

TingPing:
I guess that works, it doesn't help the situation for translators though.

PR #2327, opened 2019-04-21
Update servlist.c - Fixes #2324

### LINKED ISSUE #2324: Furnet should not be in the default network list
I don't think it is appropriate for a cartoon bestiality porn community to be advertised in Hexchat. I think I am an open-minded person when it comes to these things and what consenting adults want to do in their private lives is their business, but there's no need for bizarre sexual fetish networks to be included in the network selection dialog.

jwshields:
Update to servlist.c, removes 3 lines for furnet; fixes #2324


SoniEx2:
:<

PR #2341, opened 2019-05-21
Make dcc_ip being a per-server value.

brozs:
Moved dcc_ip from prefs to sess->server.

Currently the DCC IP address is a global setting in Preferences. It can be set globally, and optionally, the IP address can be acquired from the IRC server. Currently only a single IP address can be stored, the last detected IP address wins.

When using multiple IRC servers or networks from which at least one is available via VPN while the other use Internet, only the last detected IP address is stored, causing DCC not working with autodetected IP address on the first-connected IRC server or network.

This change makes dcc_ip a server property, allowing per-server record of the detected address. Very useful when both "VPN" and "Internet" connections use dynamic addresses.

The global setting still overrides it, if set.

### DIFF
--- a/src/common/dcc.c
+++ b/src/common/dcc.c
@@ -1618,13 +1618,13 @@ dcc_accept (GIOChannel *source, GIOCondition condition, struct DCC *dcc)
 }
 
 guint32
-dcc_get_my_address (void)	/* the address we'll tell the other person */
+dcc_get_my_address (session *sess)	/* the address we'll tell the other person */
 {
 	struct hostent *dns_query;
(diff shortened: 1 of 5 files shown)


brozs:
The LGTM check failed appears not being my fault :).

TingPing:
Looks and sounds good, I'm busy atm but will try to test this next week.

TingPing:
Also if you rebase on master I believe the LGTM checks should pass.

brozs:
Thank you, LGTM analysis now runs fine.

Issue #569, opened 2013-04-30
Add extra text field for SASL username


Nothing4You:

sugudega:
I use a different username than my registered nickname, in combination with restricting access to my nickname, to this username.

While usernames of course aren't private, this prevents random people trying to use my nickname, being matched by my access list (if I only match the domain part of my reverse DNS, and we are on the same domain).

Using the username for SASL is overloading its meaning anyway, and prevents proper Ident if you don't control your Ident daemon, and you cannot, or do not want to use your computer login, as your nickname on IRC.

Please at least add a new login method as "SASL (nickname + password)", which would use the main configured nickname, even if currently in use by someone else (well, SASL happens before declaring the desired nickname, right? so this would only be a problem for people who want to use a different nickname than the one they login with... but if they are grouped nick, wouldn't they be able to login with SASL with this other nickname too?).

PR #2345, opened 2019-05-25
Implement SASL setting that uses nick as username

vi:
Just like https://github.com/hexchat/hexchat/issues/569#issuecomment-300791905 suggested.

Partially addresses #569.

This new mode is automatically enabled for Freenode by default.


TingPing:
I strongly dislike this approach. I think the dialog should just grow a SASL username field. It had one originally but the previous maintainer removed it for cleanliness. I personally believe that experiment didn't work out and its resulted in a worse UX.

TingPing:
So I'm just going to close this. New PR welcomed.

PR #2348, opened 2019-06-02
Add 6697 to server list

WheresTheSugar:
Added 6697 into server list


TingPing:
This network doesn't seem to be up anymore?

PR #2376, opened 2019-09-07
Russian translation update

x61s:
A couple of LHFs from main windows and main connection dialog.


TingPing:
Translations are handled on Transifex: https://www.transifex.com/hexchat/

PR #2387, opened 2019-10-11
add vscode config folder to ignore list

tvieira:
very minor change to include .vscode into the ignore list.


TingPing:
This is probably something you should just add to your global git ignore rather than every project that exists.

PR #2403, opened 2019-12-15
Add include for winuser.h for mingw compilers.

ZachBacon:
Without the include gcc will complain about WM_TIMECHANGE as undeclared.

### DIFF
--- a/src/fe-gtk/maingui.c
+++ b/src/fe-gtk/maingui.c
@@ -53,6 +53,10 @@
 #include "sexy-spell-entry.h"
 #include "gtkutil.h"
 
+#ifdef WIN32
+#include <winuser.h>
+#endif
+
 #define GUI_SPACING (3)
 #define GUI_BORDER (0)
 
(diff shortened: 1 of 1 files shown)


src/fe-gtk/maingui.c
```diff
@@ -53,6 +53,10 @@
 #include "sexy-spell-entry.h"
 #include "gtkutil.h"
 
+#ifdef WIN32
```
TingPing:
```suggestion
#ifdef G_OS_WIN32
```

ZachBacon: Sure, I was just basing on some of other defines I found in the project already. If you want besides that, I can look over the code and change the Win32 defs to the glib provided ones.

TingPing: The place `WM_TIMECHANGE` is used is protected by `G_OS_WIN32` so might as well match there.

Don't bother changing others, its just churn.

ZachBacon:
That was a bit premature to close the PR, I'm discovering other issues, most likely another specific windows include.

TingPing:
I imagined you tested it. It would be nice to get Github Actions working with mingw to test these things.

ZachBacon:
@tingping I did, but then I ended up working on another set of changes which honestly I should of did a branch on (thankfully didn't commit them) so I was able to retest, what was needed was windows.h instead.

PR #2404, opened 2019-12-16
This is to correct my errant include.

ZachBacon:
MinGW needs windows.h not winuser.h, while yes winuser.h has the declaration, windows.h has the proper requirements needed and includes winuser.h and allows hexchat to compile against MinGW-w64 Version 6+


TingPing:
needs rebase (and choose a better commit message)

PR #2408, opened 2019-12-20
Adding appveyor.yml

prawnsalad:
Untested, but the formatted yaml appears to match what is expected.


TingPing:
Ended up switching to Github Actions.

PR #2413, opened 2019-12-30
Fix incorrect translations such as なると

titsuki:
Hi, I found some incorrect translations.

For example: 
+ "なると" means food (c.f., https://ja.wikipedia.org/wiki/鳴門巻き) and doesn't make sense in a context of IRC.
+ "略奪" (Note: Maybe mistakenly "奪略" was used) is often used in a context of war but not used in a context of a privilege control feature.

### DIFF
--- a/po/ja_JP.po
+++ b/po/ja_JP.po
@@ -195,19 +195,19 @@ msgstr "オペレーターアクション(_P)"
 
 #: src/common/hexchat.c:892
 msgid "Give Ops"
-msgstr "なるとを与える"
+msgstr "オペレーター権限を与える"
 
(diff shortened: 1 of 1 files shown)


TingPing:
All translations are handled here: https://www.transifex.com/hexchat/

titsuki:
@TingPing Thanks for your reply! I didn't noticed such a page by my oversight.

TingPing:
If you make an account there just let me know and I'll add you to the language where you can just make any changes.

titsuki:
@TingPing Sorry for the late reply. I already sent a request to join the team about two days ago.

TingPing:
Added

PR #2418, opened 2020-01-07
Remove more bestality focused networks

### LINKED ISSUE #2324: Furnet should not be in the default network list

driveby-hexchat:
This is an extension of #2327 which was a patch to fix #2324. In that issue it became apparent there were more server located in the recommended server list which focus on _animated bestiality_.

This patch fixes this by removing those offending servers from the recommended list.

### DIFF
--- a/src/common/servlist.c
+++ b/src/common/servlist.c
@@ -80,9 +80,6 @@ static const struct defaultserver def[] =
 	{0,			"irc.kis.lt"},
 	{0,			"irc.vub.lt"},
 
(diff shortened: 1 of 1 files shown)


TingPing:
I'm not a fan of people making fake accounts to make drive-by drama.

PR #2433, opened 2020-02-08
Fetch latest .po files

titsuki:
Hi,
All of the translations are outdated compared to the ones at transifex.
So, I imported the latest ones listed in `po/LINGUAS` via transifex API as follows:
```
#!/usr/bin/bash

TOKEN=api:<MY SECRET API TOKEN>
BASE_URL=https://www.transifex.com/api/2/project/hexchat/resource/main/translation
cat LINGUAS | xargs -n 1 -i bash -c "curl -L --user $TOKEN -X GET ${BASE_URL}/{}/?file > {}.po"
```


titsuki:
@TingPing Thanks for your review!

TingPing:
btw transifex makes a cli tool you can use for this.

titsuki:
Thanks for your advice!

To tell the truth, I already knew the cli tool because this API is one of the undocumented but public APIs: https://github.com/transifex/transifex-client/blob/master/txclib/urls.py
But at that time, it's easier to call the API via curl to me since I didn't have pip.
I'll use the official cli tool next time.

PR #2446, opened 2020-03-29
Fixed typo HexChar -> HexChat

tapwag:


TingPing:
Translations are handled on [Transifex](https://www.transifex.com/hexchat/hexchat/dashboard/)

This was simple so I fixed it there.

PR #2452, opened 2020-04-12
Upgrade ms build system

DjLegolas:
This change contains:
* Switch to Visual Studio 2019 (PlatformToolset set to v142).
* Inno path will now will be looked from `Program Files` folder directly.
* `platform.winmd` and `windows.winmd` are now being looked at the correct location.
* .NETFramework is now 4.6.1.

Note:
AppVeyor must be switched to Visual Studio 2019 for the build to work:
https://ci.appveyor.com/project/DjLegolas/hexchat/builds/32124375

### DIFF
--- a/.github/workflows/ubuntu-build.yml
+++ b/.github/workflows/ubuntu-build.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           fetch-depth: 1
 
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -44,11 +44,11 @@ jobs:
 
           New-Item -Path "c:\gtk-build" -Name "python-2.7" -ItemType "Directory"
           New-Item -Path "c:\gtk-build" -Name "python-3.6" -ItemType "Directory"
-          New-Item -Path "c:\gtk-build\python-2.7" -Name "${{ matrix.platform }}" -ItemType "SymbolicLink" -Value "C:/hostedtoolcache/windows/Python/2.7.17/${{ matrix.arch }}"
+          New-Item -Path "c:\gtk-build\python-2.7" -Name "${{ matrix.platform }}" -ItemType "SymbolicLink" -Value "C:/hostedtoolcache/windows/Python/2.7.18/${{ matrix.arch }}"
           New-Item -Path "c:\gtk-build\python-3.6" -Name "${{ matrix.platform }}" -ItemType "SymbolicLink" -Value "C:/hostedtoolcache/windows/Python/3.6.8/${{ matrix.arch }}"
 
           C:/hostedtoolcache/windows/Python/3.6.8/${{ matrix.arch }}/python.exe -m pip install cffi
-          C:/hostedtoolcache/windows/Python/2.7.17/${{ matrix.arch }}/python.exe -m pip install -qq cffi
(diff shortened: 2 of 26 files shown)


TingPing:
Nice, I was looking into this but hit the notifications issue. Instead of upgrading appveyor can you look into rebasing onto this and using Github Actions: https://github.com/hexchat/hexchat/commits/github-actions

TingPing:
Also we will need to rebuild all dependencies with VS 2019 also: https://github.com/wingtk/gvsbuild

DjLegolas:
> Nice, I was looking into this but hit the notifications issue. Instead of upgrading appveyor can you look into rebasing onto this and using Github Actions: https://github.com/hexchat/hexchat/commits/github-actions

It took me 50 tries, but I finally managed to run a complete run, on both x86 and x64 and create the artifacts.
This build system is 2 times slower (take 4 minutes for each build run, compared to ~2 minutes with appveyor) and the artifact for the build directory is about 4 times the size (~120MB compared to ~30MB).
Also, I changed the installer file name as space is an invalid character in an artifact name.

> Also we will need to rebuild all dependencies with VS 2019: https://github.com/wingtk/gvsbuild

How do you want to do it?
I'll try to run it on my local machine with VS2019.

TingPing:
I'm not too worried about a few min of build time, regarding the artifacts maybe we can compress the dir.

> How do you want to do it?
> I'll try to run it on my local machine with VS2019.

So this is where things get messy.

Firstly we need to rebase our custom patches. It has been a very long time so I *think* this is the correct branch: https://github.com/hexchat/gvsbuild/commits/hexchat-fixes-2

Secondly build everything hexchat specifically needs. I'm not sure Its documented anywhere but I believe it is just `openssl gtk2 lgi` or along those lines. You can see https://dl.hexchat.net/gtk/ for the expected contents and you can just upload your results somewhere.

Thirdly if you want to be a true hero you could get gvsbuild automatically building what we need with CI, ideally with Github Actions, or by extending upstreams support (https://github.com/wingtk/gvsbuild/blob/master/azure-pipelines.yml).

This is probably a decent chunk of work if you don't want to commit to that but I'd gladly pay for the CI work if you are interested.

DjLegolas:
> Firstly we need to rebase our custom patches. It has been a very long time so I think this is the correct branch: https://github.com/hexchat/gvsbuild/commits/hexchat-fixes-2 (they can be upstreamed too, they weren't because I think the patches were specific to a single VS version)

Done.

> Secondly build everything hexchat specifically needs. I'm not sure Its documented anywhere but I believe it is just `openssl gtk2 lgi`. You can see https://dl.hexchat.net/gtk/ for the expected contents and you can just upload your results somewhere. And of course test hexchat against it.

Well, the new build requires `gtk3 openssl lgi enchant`. And because of the change from GTK2 to GTK3 there are errors which must be taken care of.

> Thirdly if you want to be a true hero you could get gvsbuild automatically building what we need with CI, ideally with Github Actions, or by extending upstreams support (https://github.com/wingtk/gvsbuild/blob/master/azure-pipelines.yml).

Started working on it on GitHub's workflow. But although there are no errors with compiling locally, the workflows always fail... Currently I'm working on compiling `cairo` without any errors...

TingPing:
> Well, the new build requires gtk3 openssl lgi enchant. And because of the change from GTK2 to GTK3 there are errors which must be taken care of. On hold because of the next section...

HexChat doesn't use GTK3 :)

DjLegolas:
I know, but `gvsbuild` currentlly supports GTK3 only.
Do you want to upgrade or not?

TingPing:
Of course gvsbuild supports GTK2, I guess its called `gtk` not `gtk2` as I commented.

TingPing:
This was merged into master. Though our gtk builds still need updated.

PR #2458, opened 2020-04-19
win32: Replace include of winuser.h with windows.h

delthas:
winuser.h should never be included directly. windows.h should be included instead.

This fixes a critical build issue added in c5d47fc which makes all MinGW builds fail.

See #2403.


jeanthom:
Thank yu sir for your involvement in hexchats projects .

zapashcanon:
Glad this was merged.

PR #2459, opened 2020-04-19
win32: Fix building executables with invalid entrypoints

delthas:
Windows builds of the GTK frontend use the pie flag to compile
hexchat.exe. Windows needs an explicit entrypoint when compiling with
--pie, otherwise an invalid executable is created.

This sets the entrypoint of the executable on Windows (as it is
currently set in the Visual Studio project files).

This fixes a critical build issue which prevents all Windows builds using
Meson from working.


delthas:
some context: https://sourceforge.net/p/mingw-w64/mailman/message/31034877/

PR #2460, opened 2020-04-19
win32: Fix undefined symbol for builds with -with-plugin=false

delthas:
Windows builds without plugins can use notification-windows.c, which
uses module_load in its notification_backend_init function.

module_load was previously guarded with a USE_PLUGIN ifdef, but we do
need this function for Windows builds even if plugins are disabled.

This fixes a critical build issue for all Windows builds without
plugins.

### DIFF
--- a/src/common/plugin.c
+++ b/src/common/plugin.c
@@ -355,7 +355,8 @@ plugin_kill_all (void)
 	}
 }
 
-#ifdef USE_PLUGIN
+#if defined(USE_PLUGIN) || defined(WIN32)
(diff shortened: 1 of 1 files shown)


TingPing:
Should probably be a windows specific change then?

delthas:
OK. Updated.

src/common/plugin.c
```diff
@@ -355,7 +355,7 @@ plugin_kill_all (void)
 	}
 }
 
-#ifdef USE_PLUGIN
```
TingPing, line 358: Add comment that its used in `notification-windows.c`

delthas:
OK, updated.

PR #2461, opened 2020-04-19
win32: Disable ASLR for Windows debug builds

delthas:
GDB is usually able to debug executables with ASLR by temporarily
disabling ASLR when running that executable. This is only supported on
Linux. On Windows, GDB cannot debug ASLR executables.

This removes the dynamicbase linker flag on Windows for debug builds in
order to be able to debug that executable later.

Hardening an executable with ASLR is important for release builds, but
for debug builds being able to debug is much more important.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -134,9 +134,11 @@ test_ldflags = [
   '-Wl,-z,relro',
   '-Wl,-z,now',
   # mingw
-  '-Wl,--dynamicbase',
   '-Wl,--nxcompat',
 ]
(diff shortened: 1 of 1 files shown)


delthas:
some context: https://github.com/rust-lang/rust/issues/17098

delthas:
This commit does remove ASLR from debug and debugoptimized Windows builds. Pre-built Windows binaries available on the downloads page are release builds, not debug builds, right?

PR #2465, opened 2020-05-08
Update servlist.c for the split of irc.2ch.net

EmmanuelChanel:
2ch is splitted. Now it become 3 networks: irc.2ch.net (5chirc) , irc.2ch.sc, and irc.juggler.jp . It's unknown how irc.2ch.sc is going. So I deleted it.


TingPing:
Its 2020, do they really not support TLS?

EmmanuelChanel:
> Its 2020, do they really not support TLS (or SASL for that matter)?

No, they don't really do. If yes, that is not announced.

PR #2470, opened 2020-05-21
Added irc.juggler.jp , a big Japanese IRC network.

EmmanuelChanel:
irc.juggler.jp is a big Japanese IRC network. It uses an old server software with enabling Kanji channel names in iso-2022-jp encoding.


TingPing:
I didn't add it on purpose. For one it was small and their network is clearly stuck in the 90s.

HexChats network list doesn't need to contain everything that exists.

PR #2474, opened 2020-05-29
Add "DarkScience" to default server list.

dijit:
According to [netsplit.de's top 100 list](https://netsplit.de/networks/top100.php) "DarkScience" is ranked 66 for number of users.

Notably hosting the [Hak5](https://hak5.org) and [joinmarket](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/293/files#diff-4f3c2beacd536175f30616463ddaeb33) communities, along with the social aspect of the network itself.

The environment surrounding the network is one of good standing and has a fair reputation of legal and ethical behaviour.

I think this is a fair addition to the list, please let me know if there are any concerns. :)

### DIFF
--- a/src/common/servlist.c
+++ b/src/common/servlist.c
@@ -119,7 +119,7 @@ static const struct defaultserver def[] =
 
 	{"DarkMyst", 0, 0, 0, LOGIN_SASL, 0, TRUE},
 	{0,			"irc.darkmyst.org"},
-	
+
 #ifdef USE_OPENSSL
 	{"darkscience", 0, 0, 0, LOGIN_SASL, 0, TRUE},
 	{0,			"irc.darkscience.net"},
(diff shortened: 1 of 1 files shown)


src/common/servlist.c
```diff
@@ -119,6 +119,11 @@ static const struct defaultserver def[] =
 
 	{"DarkMyst", 0, 0, 0, LOGIN_SASL, 0, TRUE},
 	{0,			"irc.darkmyst.org"},
+	
+	{"darkscience", 0, 0, 0, LOGIN_SASL, 0, TRUE},
```
TingPing, line 123: Wrap in `#ifdef USE_OPENSSL` like some of the other networks.

Fix whitespace to match others.

dijit: Thanks for the feedback, I am not sure how the whitespace got messed up (I was editing using github web).

I have made the changes you suggested, the entry will be completely removed if openssl is not available, as the network does not allow plaintext. I hope that is ok.

TingPing: You are mixing spaces when the file uses tabs.

dijit: File seems to use a mixture of tabs and spaces here:

image

I have changed the style to conform to those above.

dijit: wait, nevermind, I must have broken the file myself somehow. I will fix.

TingPing: You reformatted the entire file...

dijit: Yeah, I saw. Fixed immediately in https://github.com/hexchat/hexchat/commit/46fd9a460508f019b2be13edd616ef91432bd66a

PR #2347, opened 2019-05-30
FiSHLiM: Support for CBC mode + more commands

BakasuraRCE:
I have been working to provide to hexchat a competent fish plugin with other IRC clients

Features added:
- CBC mode on SETKEY and KEYX commands
- Store keys in CBC mode (addon_fishlim.conf)
- Detect context in DELKEY command
- Encrypted flag for incoming and outgoing messages

New cipher core:
- Use of openssl EVP functions instead direct openssl functions (Now only hexchat will have this)

How to use the CBC mode:

- SETKEY command
  - for CBC mode -> /setkey cbc:key
  - for ECB mode -> /setkey ecb:key
  - for ECB mode -> /setkey key

- KEYX command: nothing for your part!

This merge keep backward compatibility with the database file addon_fishlim.conf

Standalone version: https://github.com/BakasuraRCE/hexchat-fishlim-reloaded

PR #2483, opened 2020-07-10
Use latest meson version

BakasuraRCE:
Dear @TingPing,

I'm currently developing the [tests](https://github.com/BakasuraRCE/hexchat/blob/a41918b0a7ff41a8f9b94471973fabbe8e899d33/plugins/fishlim/tests/tests.c) and [features](https://github.com/BakasuraRCE/hexchat-fishlim-reloaded/issues/1) of [Fishlim](https://github.com/hexchat/hexchat/pull/2347) plugin, for it I use the [cmocka](https://cmocka.org/) testing framework that require meson 0.48.0+, but the Github Actions build is using 0.45.0.

This PR update the build to run always the latest version of meson. Could you give me feedback about if I'm in a good choice or I need to change the system of tests?

Regards.

### DIFF
--- a/.github/workflows/ubuntu-build.yml
+++ b/.github/workflows/ubuntu-build.yml
@@ -12,8 +12,9 @@ jobs:
       - name: Install Dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


TingPing:
I'd rather not add a dependency on `cmocka` tbh.

We depend on GLib and you can use that: https://developer.gnome.org/glib/stable/glib-Testing.html

PR #2486, opened 2020-07-21
SASL EXTERNAL doesn't necessitate a certificate

jesopo:
there's 2 reasons for this
- SASL `EXTERNAL` could be something other than certificate
- not even trying SASL `EXTERNAL` when there's no cert means there's no SASL failure in logs


TingPing:
What is something other than a cert that works and hexchat supports?

n1ce-0ne:
> What is something other than a cert that works and hexchat supports?
jumping in blind here but... sasl works fine, been using it for a while

jesopo:
> What is something other than a cert that works and hexchat supports?

i think that's beside the point here. this is more correct by the protocol and allows for a server to tell you that you are attempting SASL `EXTERNAL` without providing the information it wants to use to verify you, usually a client certificate.

the concept of "SASL EXTERNAL" is that the server decides which external information to use for authentication. it could be host-based, it could be which port you are connecting on, it could be absolutely anything.

PR #2488, opened 2020-07-25
Update ru.po

Svetlana-T:
consistency


TingPing:
All translations are handled on Transifex not here: https://www.transifex.com/hexchat/hexchat/

PR #2500, opened 2020-09-03
Use pango_font_metrics_get_height() to calculate font height

### LINKED ISSUE #2449: Underscores not rendered in Monospace 11/12 font
I am finding that underscores don't render, appearing as spaces, when using the default Monospace 11 or Monospace 12 font on Ubuntu 20.04 (Focal, development beta, but due to be released this month). This occurs in both the package of 2.14.3-3 and the current master branch as of today.

Increasing to Monospace 13 causes them to appear and they work in some other fonts at various sizes. I suspect most likely that the calculated bounding box for the row is being incorrectly calculated somehow to be too short.

I imagine most likely this is either in xtext or an interaction with one of it's rendering dependencies but I'm not really sure where to start on debugging it. I'm happy to do any testing or even work towards solving it but somewhere to start would be great!

jlevon:
With newer versions of Pango and some fonts, the bottom of the
text is clipped (making underscores invisible). Use the correct
calculation when needed to avoid this.

Based on an original fix by Jeff Teunissen.

Closes #2449.

### DIFF
--- a/src/fe-gtk/xtext.c
+++ b/src/fe-gtk/xtext.c
@@ -291,6 +291,9 @@ backend_font_open (GtkXText *xtext, char *name)
 	 */
 #if PANGO_VERSION_CHECK(1, 44, 0)
 	xtext->fontsize = pango_font_metrics_get_height (metrics) / PANGO_SCALE + 1;
+
+	if (xtext->fontsize == 0)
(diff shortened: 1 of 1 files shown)


jlevon:
I tested this under Ubuntu 20.04 with a font that demonstrated the problem and verified it looks good again.

PR #2511, opened 2020-10-06
Add 6697 to default server list

MrReeee:
This network has been around for a couple of years and would like to be added to the default server list for hexchat.


TingPing:
Server doesn't seem to be up anymore here.

PR #2522, opened 2020-11-05
Update servlist.c

Xandrah:
Added DeltaPool to IRC Networks

### DIFF
--- a/src/common/servlist.c
+++ b/src/common/servlist.c
@@ -386,7 +386,7 @@ static const struct defaultserver def[] =
 	{"Xertion", 0, 0, 0, LOGIN_SASL, 0, TRUE},
 	{0,			"irc.xertion.org"},
 	
-	{"DeltaPool",		0},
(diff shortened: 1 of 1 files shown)


src/common/servlist.c
```diff
@@ -359,6 +359,9 @@ static const struct defaultserver def[] =
 
 	{"Xertion", 0, 0, 0, LOGIN_SASL, 0, TRUE},
 	{0,			"irc.xertion.org"},
+	
+	{"DeltaPool",		0},
```
TingPing, line 363: Update this to include that it supports SASL and SSL like other entries.

TingPing, line 363: Update this to include that it supports SASL and SSL like other entries.

TingPing (changes requested):

Xandrah (commented):
Updated to support SASL

src/common/servlist.c
Xandrah: Done!

PR #2524, opened 2020-11-09
Add ACN IRC Network

dimitrisdm:
Website: https://irc.acn.gr

Round-Robin DNS: global.acn.gr

Ports: 6667 - 6697(ssl only).

### DIFF
--- a/src/common/servlist.c
+++ b/src/common/servlist.c
@@ -58,7 +58,7 @@ static const struct defaultserver def[] =
 	/* Self signed */
 	{0,			"irc.accessirc.net"},
 
-	{"ACN", 0},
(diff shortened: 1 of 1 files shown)


src/common/servlist.c
```diff
@@ -58,6 +58,9 @@ static const struct defaultserver def[] =
 	/* Self signed */
 	{0,			"irc.accessirc.net"},
 
+	{"ACN", 0},
```
TingPing, line 61: Update this to include SASL and SSL like other entries.

TingPing (changes requested):

src/common/servlist.c
dimitrisdm: Done

PR #2544, opened 2021-03-01
url.c: add gemini & gopher parsing

mikejzx:
This PR adds URL parsing support for two "small-Internet" protocols, [Gemini](https://gemini.circumlunar.space/) and [Gopher](https://en.wikipedia.org/wiki/Gopher_(protocol)). This is especially useful for those of us who discuss these particular protocols over IRC!

Example URLs:
* gemini://gemini.circumlunar.space
* gopher://sdf.org


MasterQ32:
Nice to see you here, @mikejzx! I wanted to check out how to exactly solve this problem for HexChat as it's annoying for me as well not to be able to click gemini links and now i see that you already fixed it!

Issue #2531, opened 2020-12-19
python: Regression breaking bookmarks.py


misterhsp:
Is it possible to fix bookmarks.py. It no longer works with hexchat with python 3.9 on Debian SID (unstable)
https://github.com/TingPing/plugins/blob/master/HexChat/bookmarks.py

TingPing:
Does it print any errors?

misterhsp:
Yes, of course there are errors, I forgot :)

https://pastebin.com/J3BBbfQR






Am 19.12.20 um 19:32 schrieb Patrick:

v00v:
This might be related to fix this bug. https://stackoverflow.com/questions/28583565/str-object-has-no-attribute-decode-python-3-error

misterhsp:
Has anyone figured out where the problem is?

Thanks...

Sbgodin:
The problem is solved by my side. I've just submitted a pull request for my proposal to be checked.

You can do the same by :

1. Openning `/usr/lib/x86_64-linux-gnu/hexchat/python/_hexchat.py` (under Debian)
2. Go to the line 322, and confirm that the line is `return __decode(prefs_str).split(',')`
3. Replace the content of the line 322 by `return __decode(ffi.string(prefs_str)).split(',')`

It's rather simple as it is just about calling `ffi.string` to cast the object so it becomes a string which supports `__decode`.

I guess the problem is/was on `_hexchat.py` which is part of the main project. The add-on seems quite good.

Sbgodin:
@TingPing Can you have a look at my patch?

misterhsp:
@Sbgodin 
I have tested it. bookmarks.py with it works again.

Thanks a lot

dublades:
@Sbgodin 
Hi,
I had the same issue and this fix worked for me too.
I just installed the plugin and it worked correctly the first times. Then a major Ubuntu release showed up (version 21.04 x64).
After the OS upgrade the plugin didn't worked and I received the same error as misterhsp.
Thank you  for your support and TingPing for this nice add.

TingPing:
This plugin has never been in a release. So any distro encountering this has manually done backports and are simply missing patches. You should contact your distros packager to get that handled.

dublades:
Maybe I can't write very well in english but I sad thank you for the fix, it worked. Now, after the upgrade on 21.04 I can use the bookmarks python plugin with hexchat. It should be a core part of this irc client imho. 
I'm not complaining about anything  👍

PR #2546, opened 2021-03-07
Update _hexchat.py

Sbgodin:
Fixes https://github.com/hexchat/hexchat/issues/2531

On line 322, `__decode` cannot work (with Python3) because `prefs_str` has no attribute `decode`. This error appears when loading the addon `bookmarks.py` if there are saved bookmarks. I suppose that it appears for any add-on getting the content of `add-on_python.conf`.

So, I guess that `ffi` can cast to string what it created. So far so good. An experimented developer should have a look at it before merging.


TingPing:
Looks good, just improve the commit message "python: Fix exception in plugin prefs" or such.

PR #2571, opened 2021-05-23
Implement support for the IRCv3 SETNAME specification.

SadieCat:

### DIFF
--- a/src/common/userlist.c
+++ b/src/common/userlist.c
@@ -130,8 +130,12 @@ userlist_add_hostname (struct session *sess, char *nick, char *hostname,
 			g_free (user->hostname);
 			user->hostname = g_strdup (hostname);
 		}
-		if (!user->realname && realname && *realname)
+		if (realname && (!user->realname || strcmp(user->realname, realname)))
+		{
+			do_rehash = TRUE;
+			g_free (user->realname);
 			user->realname = g_strdup (realname);
+		}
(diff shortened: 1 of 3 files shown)


src/common/userlist.c
```diff
@@ -130,8 +130,12 @@ userlist_add_hostname (struct session *sess, char *nick, char *hostname,
 			g_free (user->hostname);
 			user->hostname = g_strdup (hostname);
 		}
-		if (!user->realname && realname && *realname)
+		if (realname && (!user->realname || strcmp(user->realname, realname)))
```
TingPing, line 133:
```suggestion
        if (realname && g_strcmp0 (user->realname, realname) != 0)
```

TingPing: I'm not sure if the empty string check is necessarily what you expect, but it keeps previous behavior.

src/common/userlist.c
```diff
@@ -130,8 +130,12 @@ userlist_add_hostname (struct session *sess, char *nick, char *hostname,
...
+		{
+			do_rehash = TRUE;
```
TingPing, line 135: Is `do_rehash` needed? That is for elements that show up in the userlist.

SadieCat: I was going off nearby code so I thought it did but now I check it doesn't. Removed.

PR #2572, opened 2021-05-23
Implement support for the IRCv3 account-tag specification.

SadieCat:

This probably conflicts with #2571 but I can resolve that if necessary.

---

One thing that could be improved with this is unsetting the user's account if the cap is enabled but the tag isn't sent. I'm reasonably sure that this is possible but I haven't verified that all servers send the tag on all messages yet.

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -1029,7 +1029,7 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		ex[0] = '!';
 
 		account = tags_data->account && *tags_data->account ? tags_data->account : "*";
-		userlist_set_account (sess, nick, account);
+		inbound_account (serv, nick, account, tags_data);
 	}
 
 	if (len == 4)
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -1026,6 +1026,10 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		ex[0] = 0;
 		safe_strcpy (nick, word[1], sizeof (nick));
 		ex[0] = '!';
+
+		if (tags_data->account)
+			inbound_user_info (sess, NULL, NULL, NULL, NULL, nick, NULL,
+							   tags_data->account, 0xff, tags_data);
 	}
 
 	if (len == 4)
@@ -1522,6 +1526,9 @@ handle_message_tags (server *serv, const char *tags_str,
 		*value = '\0';
 		value++;
 
+		if (serv->have_account_tag && !strcmp (key, "account"))
+			tags_data->account = g_strdup (value);
+
 		if (serv->have_server_time && !strcmp (key, "time"))
 			handle_message_tag_time (value, tags_data);
 	}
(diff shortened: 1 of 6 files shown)


src/common/proto-irc.c
```diff
@@ -1613,9 +1620,16 @@ irc_inline (server *serv, char *buf, int len)
...
 
 xit:
+	message_tags_data_free (&tags_data);
 	g_free (pdibuf);
 }
 
+void
+message_tags_data_free (message_tags_data *tags_data)
+{
+	g_free (tags_data->account);
```
TingPing, line 1630:
```suggestion
	g_clear_pointer (&tags_data->account, g_free);
```

TingPing:
> One thing that could be improved with this is unsetting the user's account if the cap is enabled but the tag isn't sent. I'm reasonably sure that this is possible but I haven't verified that all servers send the tag on all messages yet.

Based on the spec I'm reading this seems like a requirement.

> The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG, MODE, NOTICE, and all others). The tag also MUST be added by the ircd to all numerics directly caused by the sender. 

and:

> (If the user is not identified to any services account, the tag MUST NOT be sent.)

So as-is hexchat is left in an incorrect state.

SadieCat:
Should be done now. Thanks for the review.

src/common/proto-irc.c
```diff
@@ -1026,6 +1027,9 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		ex[0] = 0;
 		safe_strcpy (nick, word[1], sizeof (nick));
 		ex[0] = '!';
+
+		account = tags_data->account && *tags_data->account ? tags_data->account : "*";
+		userlist_set_account (sess, nick, account);
```
TingPing, line 1032: If you look at `inbound_account()` what it does is call `userlist_set_account()` for ever session in a server, which I think is what we want.

TingPing: Also looking at it, `userlist_set_account()` always frees and copies the new account, ignoring if its unchanged, so that should be handled somewhere.

SadieCat: Fixed.

src/common/userlist.c
```diff
@@ -101,13 +101,17 @@ userlist_set_account (struct session *sess, char *nick, char *account)
 	user = userlist_find (sess, nick);
 	if (user)
 	{
-		g_free (user->account);
-			
 		if (strcmp (account, "*") == 0)
+		{
+			g_free (user->account);
 			user->account = NULL;
```
TingPing, line 107:
```suggestion
			g_clear_pointer (&user->account, g_free);
```

src/common/userlist.c
```diff
@@ -101,13 +101,17 @@ userlist_set_account (struct session *sess, char *nick, char *account)
...
-		else
+		}
+		else if (g_strcmp0 (user->account, account))
```
TingPing, line 109:
```suggestion
		} else if (g_strcmp0 (user->account, account))
```

src/common/proto-irc.c
```diff
@@ -1026,6 +1027,9 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		ex[0] = 0;
 		safe_strcpy (nick, word[1], sizeof (nick));
 		ex[0] = '!';
+
```
TingPing, line 1030: Being in this else block seems unclear. This block is about copying nick values.

I assume the reason is because the other block is server messages, but maybe this can be pulled out or at least commented why its here.

SadieCat: I moved it out. Now that I think about it there should be a `serv->have_account_tag` check around it anyway.

PR #2573, opened 2021-05-23
Add a workaround for icons not scaling right on HiDPI screens.

SadieCat:
Currently on Linux (Ubuntu 21.04 in my case) HiDPI screens icons are not rendered at the correct size. This is problematic if you have a HiDPI screen because icons are just completely invisible. Unfortunately because HexChat uses GTK2 it's not really possible to automatically scale to the screen resolution so this patch adds a workaround by having it check a common environment variable used for UI scaling in GDK3 and then scaling the icons by that size on load.

Without this patch on my system:

Screenshot from 2021-05-23 07-44-08

With this patch and GDK_SCALE=3 set in the environment on my system:

Screenshot from 2021-05-23 07-43-27

If anyone can recommend a better way of doing this then I'll be happy to change this.

### DIFF
--- a/src/fe-gtk/pixmaps.c
+++ b/src/fe-gtk/pixmaps.c
@@ -111,8 +111,8 @@ load_pixmap (const char *filename)
 		iscale = atoi (scale);
 		if (iscale >= 0)
 		{
-			scaledpixbuf = gdk_pixbuf_scale_simple (pixbuf, gdk_pixbuf_get_height (pixbuf) * iscale,
-				gdk_pixbuf_get_width (pixbuf) * iscale, GDK_INTERP_BILINEAR);
+			scaledpixbuf = gdk_pixbuf_scale_simple (pixbuf, gdk_pixbuf_get_width (pixbuf) * iscale,
+				gdk_pixbuf_get_height (pixbuf) * iscale, GDK_INTERP_BILINEAR);
 
 			g_object_unref (pixbuf);
 			pixbuf = scaledpixbuf;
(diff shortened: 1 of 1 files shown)


src/fe-gtk/pixmaps.c
```diff
@@ -102,6 +104,21 @@ load_pixmap (const char *filename)
 		g_free (path);
 	}
 
+	// Hack to avoid unbearably tiny icons on HiDPI screens.
+	scale = getenv ("GDK_SCALE");
```
TingPing, line 108:
```suggestion
	scale = g_getenv ("GDK_SCALE");
```

src/fe-gtk/pixmaps.c
```diff
@@ -102,6 +104,21 @@ load_pixmap (const char *filename)
...
+	if (scale)
+	{
+		iscale = atoi (scale);
+		if (iscale >= 0)
```
TingPing, line 112: `0` sounds like a pretty broken scale.

src/fe-gtk/pixmaps.c
```diff
@@ -102,6 +104,21 @@ load_pixmap (const char *filename)
...
+		{
+			scaledpixbuf = gdk_pixbuf_scale_simple (pixbuf, gdk_pixbuf_get_width (pixbuf) * iscale,
```
TingPing, line 114: `NULL` isn't handled.

src/fe-gtk/pixmaps.c
SadieCat: Ooh yeah that's supposed to be >.

PR #2574, opened 2021-05-23
Implement support for the IRCv3 invite-notify specification.

SadieCat:

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -1165,13 +1165,14 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		case WORDL('I','N','V','I'):
 			if (ignore_check (word[1], IG_INVI))
 				return;
-			
-			if (word[4][0] == ':')
-				EMIT_SIGNAL_TIMESTAMP (XP_TE_INVITED, sess, word[4] + 1, nick,
-											  serv->servername, NULL, 0,
+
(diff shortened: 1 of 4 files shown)


src/common/proto-irc.c
```diff
@@ -1165,13 +1165,14 @@ process_named_msg (session *sess, char *type, char *word[], char *word_eol[],
 		case WORDL('I','N','V','I'):
 			if (ignore_check (word[1], IG_INVI))
 				return;
-			
-			if (word[4][0] == ':')
-				EMIT_SIGNAL_TIMESTAMP (XP_TE_INVITED, sess, word[4] + 1, nick,
-											  serv->servername, NULL, 0,
+
+			text = STRIP_COLON(word, word_eol, 4);
+			if (g_strcmp0 (word[3], serv->nick))
```
TingPing, line 1170: I assume this should be case-insensitive. Use `serv->p_cmp()` (not `NULL` safe).

TingPing (changes requested):

src/common/proto-irc.c
SadieCat: Fixed.

SadieCat:
The AppVeyor CI failure seems to be due to an issue unrelated to this PR.

>Error downloading remote file: One or more errors occurred.
>Inner Exception: The remote name could not be resolved: 'dl.hexchat.net'

PR #2585, opened 2021-05-28
Parse the output of the 005 numeric correctly.

SadieCat:
This implements support for the full 005 numeric syntax including negation and value escapes as defined in [draft-hardy-irc-isupport-00](https://datatracker.ietf.org/doc/html/draft-hardy-irc-isupport-00). This fixes HexChat on servers that:

1. Have unloaded a previously supported feature at runtime (e.g. unloading the monitor module in InspIRCd removing the MONITOR token).
2. Have escaped spaces in the network name (see testnet.inspircd.org for an example of this).
3. Send a value for a token where HexChat expects none (e.g. INVEX on InspIRCd &mdash; the value for this token is optional) or vice versa.

### DIFF
--- a/src/common/modes.c
+++ b/src/common/modes.c
@@ -891,7 +891,7 @@ inbound_005 (server * serv, char *word[], const message_tags_data *tags_data)
 			serv->supports_monitor = tokadding;
 		} else if (g_strcmp0 (tokname, "NETWORK") == 0)
 		{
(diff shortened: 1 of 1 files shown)


SadieCat:
@lgtm-com appears to be drunk because none of the new issues it reports are in code I have edited.

PR #2586, opened 2021-05-30
 Implement support for the IRCv3 sts specification.

SadieCat:
This has been tested to the best of my abilities but there may still be some edge cases as the connection logic is kinda gnarly.

### DIFF
--- a/src/common/inbound.c
+++ b/src/common/inbound.c
@@ -1829,6 +1829,8 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
 					stsvalue = atoi (stsvaluestr);
 					if (stsvalue > 0)
 						sts_update_expiry(serv->hostname, stsvalue, serv->stsprofile);
+					else
+						g_debug ("Malformed STS profile duration: %s", stsvaluestr);
 				}
 			} else
 			{
@@ -1840,7 +1842,7 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
 					{
 						/* We are connecting on plain text and have found a valid STS profile. */
 						serv->stsprofile = sts_new ();
-						strcpy (serv->stsprofile->host, serv->hostname);
+						g_strlcpy (serv->stsprofile->host, serv->hostname, sizeof (serv->stsprofile->host));
 						serv->stsprofile->port = stsvalue;
 
 						/* Reconfigure with the new port and security settings. */
@@ -1859,6 +1861,8 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
 						prefs.hex_net_reconnect_delay = stsvalue;
 						break;
 					}
+					else
+						g_debug ("Malformed STS profile port: %s", stsvaluestr);
 				}
 			}
 
--- a/src/common/inbound.c
+++ b/src/common/inbound.c
@@ -47,6 +47,7 @@
 #include "ctcp.h"
 #include "hexchatc.h"
 #include "chanopt.h"
+#include "sts.h"
 
 
 void
@@ -1728,6 +1729,7 @@ static const char * const supported_caps[] = {
 	"setname",
 	"invite-notify",
 	"account-tag",
+	/* "sts", Handled manually */
 
 	/* ZNC */
 	"znc.in/server-time-iso",
@@ -1778,7 +1780,9 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
 	char buffer[500];	/* buffer for requesting capabilities and emitting the signal */
 	gboolean want_cap = FALSE; /* format the CAP REQ string based on previous capabilities being requested or not */
 	char **extensions;
-	int i;
+	int i, stsvalue;
+	GHashTable *stscap;
+	const char *stsvaluestr;
 
 	if (g_str_has_prefix (extensions_str, "* "))
 	{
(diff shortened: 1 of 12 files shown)


TingPing:
It may take me a while to better review this one.

SadieCat:
That's understandable. It's a big change.

SadieCat:
Rebased this to fix the few merge conflicts that existed from my other SSL change being merged.

SadieCat:
@TingPing Have you had a chance to review this yet

src/common/inbound.c
```diff
@@ -1811,6 +1815,57 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
...
+		if (!g_strcmp0 (extension, "sts"))
+		{
+			stscap = sts_parse_cap (value);
+			if (serv->use_ssl)
+			{
+				/* Retrieve the duration; if its valid then store/update it. */
+				stsvaluestr = (const char*) g_hash_table_lookup (stscap, "duration");
+				if (stsvaluestr)
+				{
+					stsvalue = atoi (stsvaluestr);
```
TingPing, line 1829: Could you add logging in these cases where parsing things fail. `g_debug ()`

src/common/inbound.c
```diff
@@ -1811,6 +1815,57 @@ inbound_cap_ls (server *serv, char *nick, char *extensions_str,
...
+			{
+				stsvaluestr = (const char*) g_hash_table_lookup (stscap, "port");
+				if (stsvaluestr)
+				{
+					stsvalue = atoi (stsvaluestr);
+					if (stsvalue > 0 && stsvalue < 65536)
+					{
+						/* We are connecting on plain text and have found a valid STS profile. */
+						serv->stsprofile = sts_new ();
+						strcpy (serv->stsprofile->host, serv->hostname);
```
TingPing, line 1843: I don't want `strcpy` anywhere. `g_strlcpy()` usually is the better fit.

src/common/sts.c
```diff
@@ -0,0 +1,151 @@
...
+	profiles = g_slist_append (profiles, profile);
+}
+
+void sts_update_expiry (const char *host, time_t newexpiry, struct sts_profile *incomplete_profile)
+{
+	struct sts_profile *profile;
+
+	profile = incomplete_profile ? incomplete_profile : sts_find (host);
+	if (profile)
+		profile->expiry = time (NULL) + newexpiry;
```
TingPing, line 150: This can be larger than `profile->expiry`, no validation is done.

src/common/sts.c
```diff
@@ -0,0 +1,151 @@
...
+{
+	char buf[256];
+	int fh;
+	struct sts_profile *profile;
+
+	fh = hexchat_open_file (STS_CONFIG_FILE, O_RDONLY, 0, 0);
+	if (fh < 0)
+		return; /* Filesystem not readable. */
+
+	while (waitline (fh, buf, sizeof buf, FALSE) != -1)
```
TingPing, line 64: Same comment that I want GIO based IO for reading files.

You can use [`g_data_input_stream_read_line()`](https://docs.gtk.org/gio/method.DataInputStream.read_line.html)

src/common/sts.c
```diff
@@ -0,0 +1,151 @@
...
+	{
+		profile = sts_new ();
+		if (sscanf (buf, "%s %u %ld", profile->host, &profile->port, &profile->expiry) != 3)
+		{
+			/* Malformed profile; drop it. */
```
TingPing, line 69: `g_debug()`

src/common/outbound.c
```diff
@@ -3447,6 +3447,10 @@ cmd_server (struct session *sess, char *tbuf, char *word[], char *word_eol[])
 		use_ssl_noverify = TRUE;
 		offset++;	/* args move up by 1 word */
 	}
+
+	/** If the server was previously used by a failed STS connection then free the incomplete profile. */
+	if (serv->stsprofile)
```
TingPing, line 3452: `g_clear_pointer()` is NULL safe

src/common/sts.c
```diff
@@ -0,0 +1,151 @@
...
+	table = g_hash_table_new (g_str_hash, g_str_equal);
+	entries = g_strsplit (cap, ",", 0);
+	for (currentry = entries; *currentry; ++currentry)
+	{
+		value = strchr (*currentry, '=');
+		if (value)
+			g_hash_table_insert (table, g_strndup (*currentry, value - *currentry), g_strdup (value + 1));
+	}
+
+	g_free (entries);
```
TingPing, line 106: Leak: `g_strfreev()`

src/common/sts.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+GHashTable *
+sts_parse_cap (const char* cap)
+{
+	char **entries, **currentry;
+	char *value;
+	GHashTable *table;
+
+	table = g_hash_table_new (g_str_hash, g_str_equal);
```
TingPing, line 97: Leak: `g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);`

TingPing (changes requested):
This is an initial review and this will take another round or so.

As I mentioned on IRC, I really dislike adding more blocking IO to this codebase. However this is common throughout hexchat so perhaps I should not block on that fact here.

A lot of little unsafe C nits I noticed so I will want to be a bit more detailed reviewing this later.

src/common/sts.c
SadieCat: How would you validate this in C? In C++ you would check if `std::numeric_limits<time_t>::max()` minus `newexpiry` is greater than the return value of `time(NULL)` but there's no TIME_MAX equivalent in C. The glib checked add functions don't work for time_t either.

SadieCat:
Hey, sorry this took so long. I've rebased against the latest master and fixed all of the review comments.

SadieCat:
If someone else wants to take this up then please feel free. I'm no longer interested in it.

PR #2587, opened 2021-05-30
Make default connection behaviours more secure

SadieCat:
1. If built with OpenSSL then tick the box for using SSL by default in the server list and use the typical SSL port in the example.
2. If connecting with SSL via a command then enable certificate validation by default. This can be opted-out of by using `-ssl-noverify` instead of `-ssl`.

### DIFF
--- a/src/fe-gtk/servlistgui.c
+++ b/src/fe-gtk/servlistgui.c
@@ -40,9 +40,9 @@
 #define SERVLIST_Y_PADDING 0			/* vertical padding in the network editor */
 
 #ifdef USE_OPENSSL
-# define DEFAULT_SERVER "irc.example.com/6697"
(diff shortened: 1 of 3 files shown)


src/fe-gtk/servlistgui.c
```diff
@@ -39,6 +39,12 @@
 #define SERVLIST_X_PADDING 4			/* horizontal paddig in the network editor */
 #define SERVLIST_Y_PADDING 0			/* vertical padding in the network editor */
 
+#ifdef USE_OPENSSL
+# define DEFAULT_SERVER "irc.example.com/6697"
```
TingPing, line 43: I think I prefer the previous default of not a real domain.

SadieCat: I changed it back.

PR #2589, opened 2021-06-01
Implement generic support for IRCv3 standard replies.

SadieCat:
Not much is really done with them currently other than printing them in a user-readable way but later specs may require special handling.

https://ircv3.net/specs/extensions/standard-replies

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -460,6 +460,18 @@ channel_date (session *sess, char *chan, char *timestr,
 								  tags_data->timestamp);
 }
 
+static int
+trailing_index(char *word_eol[])
+{
+	int index;
+	for (index = 3; index < PDIWORDS; ++index)
+	{
+		if (word_eol[index][0] == ':')
+			break;
+	}
+	return index;
+}
+
 static void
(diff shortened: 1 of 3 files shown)


src/common/proto-irc.c
```diff
@@ -460,6 +460,18 @@ channel_date (session *sess, char *chan, char *timestr,
 								  tags_data->timestamp);
 }
 
+static int
+trailing_index(char *word_eol[])
+{
+	int index;
```
TingPing, line 466: Avoid the name `index` since its in the stdlib.

src/common/proto-irc.c
```diff
@@ -460,6 +460,18 @@ channel_date (session *sess, char *chan, char *timestr,
 								  tags_data->timestamp);
 }
 
+static int
+trailing_index(char *word_eol[])
```
TingPing, line 464:
```suggestion
trailing_index(const char *word_eol[])
```

TingPing:
Looks good, I'll merge after minor nits.

SadieCat:
Should be resolved now?

PR #2592, opened 2021-06-18
Replace identify-msg support with solanum.chat/identify-msg.

SadieCat:
Now that freenode has committed suicide there's no significant users of ircd-seven. This means that there's essentially no point to supporting the ircd-seven specific identify-msg cap.

However, the successor to ircd-seven, Solanum, which is currently used by Libera Chat and will be used by OFTC in the future has a replacement for it which uses a cap instead of message prefix which removes the need do the messy character skipping logic. Cool!

### DIFF
--- a/src/common/ctcp.c
+++ b/src/common/ctcp.c
@@ -94,9 +94,6 @@ ctcp_handle (session *sess, char *to, char *nick, char *ip,
 	char outbuf[1024];
 	int ctcp_offset = 2;
 
-	if (serv->have_idmsg && (word[4][1] == '+' || word[4][1] == '-') )
-			ctcp_offset = 3;
-
 	/* consider DCC to be different from other CTCPs */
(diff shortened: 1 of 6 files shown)


src/common/ctcp.c
```diff
@@ -129,7 +126,7 @@ ctcp_handle (session *sess, char *to, char *nick, char *ip,
 		if (ctcp_check (sess, nick, word, word_eol, word[4] + ctcp_offset))
 			goto generic;
 
-		inbound_action (sess, to, nick, ip, msg + 7, FALSE, id, tags_data);
+		inbound_action (sess, to, nick, ip, msg + 7, FALSE, tags_data->identified, tags_data);
```
SadieCat, line 129: This parameter (and others) could be removed but I'm guessing that would break plugins.

PR #2596, opened 2021-06-22
Fix parsing +beI lists on InspIRCd.

SadieCat:
This fixes showing the set time of bans on ban lists.

### DIFF
--- a/src/common/proto-irc.c
+++ b/src/common/proto-irc.c
@@ -813,7 +813,7 @@ process_numeric (session * sess, int n,
 		break;
 
 	case 346:	/* +I-list entry */
-		if (!inbound_banlist (sess, atol (word[7]), word[4], word[5], word[6], 346,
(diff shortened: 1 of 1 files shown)


SadieCat:
Note: these would be much better off fixed by a standards compliant parser. I can submit a PR which switches incoming message parsing from `process_data_init` and the `WORDL` hack to a new to-spec parser if there is interest.

TingPing:
> I can submit a PR which switches incoming message parsing from process_data_init and the WORDL hack to a new to-spec parser if there is interest.

That would be great, but do note that the word arrays are public plugin API so it must still be done.

I wrote this a while back: https://github.com/TingPing/irc-client/blob/master/lib/irc-message.c

I just didn't want to change all usage in the codebase.

TingPing:
> and the WORDL hack

This may or may not be relevant also: https://github.com/TingPing/irc-client/blob/master/lib/irc-server.c#L1357-L1407

PR #2609, opened 2021-07-12
Avoid direct use of libproxy

mcatanzaro:
Since hexchat already depends on GLib, it's better to use GProxyResolver
instead. This might use libproxy, or not, as appropriate.

P.S. This removes a memory safety issue because proxy_list is allocated
using malloc(), not g_malloc(), and therefore using g_strfreev() is
incorrect. The proper way to free the proxy list returned by libproxy
is to use px_proxy_factory_free_proxies() (but nobody does that because
it was added in libproxy 0.4.16, which is somewhat recent).

Note: libproxy is not used by default. It's only used if the proxy type setting in Preferences is switched to Auto, i.e. `net_proxy_type = 5` in hexchat.conf. I didn't test hexchat with an actual proxy server (who uses proxies in 2021?) but I did test that the proxy variables in server_child() get set properly by adding a debug print:

```
diff --git a/src/common/server.c b/src/common/server.c
index c2a046e6..7afaa375 100644
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -1429,6 +1429,7 @@ server_child (server * serv)
                                *c = '\0';
                                proxy_port = atoi (c + 1);
                        }
+g_warning ("Decided to use proxy %s (%d)", proxy, proxy_type);
 
                        g_strfreev (proxy_list);
                        g_free (url);
```

By default it looks like this:

```
** (hexchat:30991): WARNING **: 08:51:06.601: Decided to use proxy direct:// (0)
```

No proxy, everything works fine. But if I sabotage my GNOME HTTP proxy settings to use http://127.0.0.1 (I didn't try an actual proxy server), then hexchat fails to connect and my debug prints this instead:

```
** (hexchat:31206): WARNING **: 08:51:35.485: Decided to use proxy http://127.0.0.1:8080 (4)
```

Seems good.

### DIFF
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -61,10 +61,6 @@
 #include "ssl.h"
 #endif
 
-#ifdef USE_LIBPROXY
-#include <proxy.h>
-#endif
-
 #ifdef USE_OPENSSL
 /* local variables */
 static struct session *g_sess = NULL;
@@ -78,9 +74,15 @@ static void server_disconnect (session * sess, int sendquit, int err);
 static int server_cleanup (server * serv);
 static void server_connect (server *serv, char *hostname, int port, int no_login);
 
-#ifdef USE_LIBPROXY
-extern pxProxyFactory *libproxy_factory;
-#endif
+static void
+write_error (char *message, GError **error)
+{
+	if (error == NULL || *error == NULL) {
+		return;
+	}
+	g_printerr ("%s: %s\n", message, (*error)->message);
+	g_clear_error (error);
+}
 
 /* actually send to the socket. This might do a character translation or
    send via SSL. server/dcc both use this function. */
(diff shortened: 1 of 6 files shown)


src/fe-gtk/setup.c
```diff
@@ -614,9 +614,7 @@ static const char *const proxytypes[] =
 	N_("SOCKS4"),
 	N_("SOCKS5"),
 	N_("HTTP"),
-#ifdef USE_LIBPROXY
 	N_("Auto"),
```
mcatanzaro, line 617: Note this setting will be broken on Windows if (a) you don't build libproxy, or (b) you don't build glib-networking with libproxy support enabled.

src/common/server.c
```diff
@@ -1392,14 +1394,17 @@ server_child (server * serv)
...
 		{
 			char **proxy_list;
 			char *url, *proxy;
+			GProxyResolver *resolver;
+			GError *error = NULL;
 
+			resolver = g_proxy_resolver_get_default ();
 			url = g_strdup_printf ("irc://%s:%d", hostname, port);
-			proxy_list = px_proxy_factory_get_proxies (libproxy_factory, url);
+			/* FIXME: do an async lookup to avoid blocking */
```
TingPing, line 1406: You can remove this, it happens in a child process that doesn't block the parent.

mcatanzaro: OK, done.

PR #2612, opened 2021-07-18
Updated translation

Felipefpl:
Updated translation to brazilian portuguese.

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -1,7 +1,7 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the hexchat package.
-# 
+#
 # Translators:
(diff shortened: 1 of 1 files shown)


TingPing:
Translations are all handled on Transifex rather than Github PRs, see https://hexchat.readthedocs.io/en/latest/helping.html#translation

Felipefpl:
Ok, thanks for the info, i signed there and i'm waiting to be added.

TingPing:
Accepted.

Felipefpl:
Thanks for that but now i have another problem, more than 93% of the translation is marked as reviewed by a previous translator who did his work more than 3 years ago, because of that i cant add the hundreds of fixes i made to the translation unless someone removes the reviewed status.

TingPing:
I've changed your permissions so you should be able to make changes.

Felipefpl:
> I've changed your permissions so you should be able to make changes.

Thank you very much for this, i went ahead and removed the "reviewed" status from the translation, i'm gonna let stay that so the translators may make their changes without any problems in the future. :)

PR #2621, opened 2021-08-22
Remove some weird guesswork on the 004 numeric.

SadieCat:
Bahamut and ircu both send 005 MODES and ELIST so this is entirely unnecessary. The other IRCd checked for here is for a dead network.

While we're editing this code fix HexChat on servers that can only support one mode at a time (these are mostly gateway servers).

### DIFF
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -1770,6 +1770,7 @@ server_set_defaults (server *serv)
 	serv->chanmodes = g_strdup ("beI,k,l");
 	serv->nick_prefixes = g_strdup ("@%+");
 	serv->nick_modes = g_strdup ("ohv");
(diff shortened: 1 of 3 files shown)


src/common/server.c
```diff
@@ -1770,6 +1770,7 @@ server_set_defaults (server *serv)
 	serv->chanmodes = g_strdup ("beI,k,l");
 	serv->nick_prefixes = g_strdup ("@%+");
 	serv->nick_modes = g_strdup ("ohv");
+	serv->modes_per_line = 3;
```
TingPing, line 1773: Could you add a comment that links to an rfc?

SadieCat: Done.

PR #2624, opened 2021-08-23
Increase the linebuf length to fit a full message including tags.

SadieCat:
In addition to the 512 bytes specified by RFC 1459 IRCv3 message tags can add up to 8191 extra bytes of data to a message. This means that the previous limit of 4096 is far too small to hold an IRC message whilst requesting caps that enable tags,

### DIFF
--- a/src/common/hexchat.h
+++ b/src/common/hexchat.h
@@ -503,7 +503,7 @@ typedef struct server
 	char servername[128];			/* what the server says is its name */
 	char password[86];
 	char nick[NICKLEN];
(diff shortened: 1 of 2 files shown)


TingPing:
My one concern, that has popped up elsewhere, is that buffer changes in one place can sometimes break assumptions later that just blindly copy into an existing smaller buffer. I think *most* of those have been removed but they still exist. If you've glanced at the usage and didn't see anything I'll merge as-is.

SadieCat:
I checked through as many places as I could and I couldn't see any places that this might cause an issue.

PR #2636, opened 2021-10-04
change joind default radio button

mapreri:
in the joind dialog window, default the radio button to "join this channel" if we have a channel to recommend

Otherwise the radio defaults to the first entry "Nothing, I'll join a
channel later", which in my opinion is weird if we are at the same time
pre-setting a channel name.


TingPing:
I did this on purpose a very long time ago.

A lot of users quickly just click past that screen and then got dumped in our support channel. This mostly lead to anger and confusion. So instead I left it be an active choice to join the channel we recommend.

mapreri:
ah, makes sense.  thanks for the history bit ;)

Issue #2507, opened 2020-09-29
Nicknames with not-ASCII symbols not recognized


kmikita:
If nickname contain not-ASCII symbols then it does not clickable in text window.

![wrong_action-1](https://user-images.githubusercontent.com/72072194/94539487-f587ff00-024d-11eb-91c3-ec5c03d14597.png)
![wrong_action-2](https://user-images.githubusercontent.com/72072194/94539484-f456d200-024d-11eb-8986-b896c3abfffc.png)

![right_action-1](https://user-images.githubusercontent.com/72072194/94539490-f6209580-024d-11eb-8b1d-3a2bbf58d20e.png)
![right_action-2](https://user-images.githubusercontent.com/72072194/94539489-f587ff00-024d-11eb-92f6-8a0d160821ed.png)

TingPing:
So there are a few possible places something could go wrong.

Firstly hovering over a word it has to detect word boundries. At a glance it should respect anything utf-8:

https://github.com/hexchat/hexchat/blob/7a275812c0002fe10db5c60e29a34ba6ce4cede1/src/fe-gtk/xtext.c#L1536


Then once it finds a word it has to compare that against usernames in the user list. This will usually work since people don't change their nicks often. However case-sensitivity is completely broken for non-ascii and could very much be entirely broken.

PR #2654, opened 2021-11-19
src/common/url.c : #2507 add UTF Cyrillic support in nicknames

kmikita:
On some servers, for example RusNET, the use of Cyrillic characters in nicknames is allowed, which leads to issue #2507 .

### DIFF
--- a/src/common/url.c
+++ b/src/common/url.c
@@ -620,7 +620,8 @@ re_email (void)
 /*	NICK description --- */
 /* For NICKPRE see before url_check_word() */
 #define NICKHYP	"-"
-#define NICKLET "a-z"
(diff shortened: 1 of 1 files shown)


TingPing:
The entire client is broken with non-ascii nicks, so I don't think it makes sense to change tiny details like this to support them.

kmikita:
Ideally, the entire range of UTF should be passed, but I was only add the Cyrillic alphabet as this solves this solves a particular annoying issue #2705. I don't see any stability or performance hindrance in the use of non-ASCII characters.

TingPing:
It can work if the server only does ascii comparisons for nicks.

PR #2672, opened 2022-01-17
Fix handling invalid ports.

SadieCat:
Instead of wrapping around, which is not behaviour any reasonable user would expect, just use the default port if above 65535.

Disallow connecting on port 0. This port has special meaning and servers can not listen on it. It is more likely the user just gave an invalid value to the port field as `atoi("invalid") == 0`.

### DIFF
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -1567,6 +1567,7 @@ server_connect (server *serv, char *hostname, int port, int no_login)
 		if (serv->use_ssl)
 			port = 6697;
 #endif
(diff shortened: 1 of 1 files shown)


src/common/server.c
```diff
@@ -1568,7 +1568,6 @@ server_connect (server *serv, char *hostname, int port, int no_login)
 			port = 6697;
 #endif
```
TingPing, line 1569: Throw a `g_debug ("Attempted to connect to invalid port, assuming default %u", port);` in there.

SadieCat: Done.

PR #2673, opened 2022-01-21
meson: Fix exported functions in plugins

Biswa96:


Biswa96:
I should have done this in previous pull request but did not notice it 🙇

Issue #2666, opened 2022-01-04
/execwrite aka cmd_execw (common/outbound.c)  echos parameters


orcus-de:
/execwrite aka cmd_execw (common/outbound.c)  echos parameters unconditionally at client textbox using PrintText

This looks like some (forgotten?) debug-output:
I would normally expect some -d (or other debug option) to explicitly echo input-data.
In case the current behavior should not be changed for some unknow reasons - as an alternative some -q (quiet) would help me too.

Background:
I'm using /exec to run external commands and want to feed in data using /execwrite (to avoid parameter expansion by the shell if being added to /exec directly).
As the run tool is used by some more users, its a bit unexpected for them to get the used parameters shown at the textbox.

orcus-de:
Idk if this would help working on the topic - I've tested my two suggested variants by modifying
common/outbound.c accordingly - where adding and handling an optional -q parameter would
keep up compatibility with the previous version, but would allow users to suppress the textbuffer
output if needed ...

```
static int
cmd_execw (struct session *sess, char *tbuf, char *word[], char *word_eol[])
{
	int len;
	char *temp;
	exec_check_process (sess);
	if (sess->running_exec == NULL)
	{
		EMIT_SIGNAL (XP_TE_NOCHILD, sess, NULL, NULL, NULL, NULL, 0);
		return FALSE;
	}

#if 0
	// variant 1 - use a switch -d to show debug output at the text buffer
	// this will change the default behavior compared to previous version
	// orcus - only echo to textbuffer if -d being used
	if ( 0==strcmp (word[2], "-d") )
	{
		len = strlen(word_eol[3]);
		temp = g_strconcat (word_eol[3], "\n", NULL);
		PrintText(sess, temp);
	}
	else
	{
		len = strlen(word_eol[2]);
		temp = g_strconcat (word_eol[2], "\n", NULL);
	}
#else
	// variant 2 - use a switch -q to quiet/suppress output at the text buffer
	// this will keep the default behavior compared to previous version
	// orcus - only suppress echo to textbuffer if -q being used
	
	if ( 0==strcmp (word[2], "-q") )
	{
		len = strlen(word_eol[3]);
		temp = g_strconcat (word_eol[3], "\n", NULL);
	}
	else
	{
		len = strlen(word_eol[2]);
		temp = g_strconcat (word_eol[2], "\n", NULL);
		PrintText(sess, temp);
	}
#endif
	write(sess->running_exec->myfd, temp, len + 1);
[...]

PR #2675, opened 2022-01-21
Add -q/-- flags to /execwrite to EXECWRITE and cmd_execW

orcus-de:
see: [https://github.com/hexchat/hexchat/issues/2666](url)

added two flags to EXECWRITE and cmd_execw
-q : (quiet) to allow suppressing of additional (debug) output at the text box
--  : (stop parsing for further flags) for the edge cases where -q itself migh be part of used data and the user wants to show that at the text box

### DIFF
--- a/src/common/outbound.c
+++ b/src/common/outbound.c
@@ -1579,9 +1579,29 @@ cmd_execw (struct session *sess, char *tbuf, char *word[], char *word_eol[])
 		EMIT_SIGNAL (XP_TE_NOCHILD, sess, NULL, NULL, NULL, NULL, 0);
 		return FALSE;
 	}
-	len = strlen(word_eol[2]);
-	temp = g_strconcat (word_eol[2], "\n", NULL);
-	PrintText(sess, temp);
+	// use -q flag to quiet/suppress output at text box
+	// use -- flag to stop interpreting arguments as flags
+	//        needed if -q itself would occur as data
+	if ( 0==strcmp (word[2], "--") )
+	{
+		len = strlen(word_eol[3]);
+		temp = g_strconcat (word_eol[3], "\n", NULL);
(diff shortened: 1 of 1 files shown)


src/common/outbound.c
```diff
@@ -1579,9 +1579,29 @@ cmd_execw (struct session *sess, char *tbuf, char *word[], char *word_eol[])
 		EMIT_SIGNAL (XP_TE_NOCHILD, sess, NULL, NULL, NULL, NULL, 0);
 		return FALSE;
 	}
-	len = strlen(word_eol[2]);
-	temp = g_strconcat (word_eol[2], "\n", NULL);
-	PrintText(sess, temp);
+	// use -q flag to quiet/suppress output at text box
+	// use -- flag to stop interpreting arguments as flags
+	//        needed if -q itself would occur as data
+	if ( 0==strcmp (word[2], "--") )
```
TingPing, line 1585: Just a style nit to change all of these to:

```c
if (strcmp (...) == 0)
```

TingPing:
Also change commit message to something more descriptive. Like *"Add -q/-- flags to /execwrite"*

orcus-de:
np - just drop the comments - they are meant for the integrator anyways

I still advocat for left sided constant numerics on comparrison expressions, as it catches typos of using = instead of the intended == by mistake, which will never compile
the other way round might (not in this case) just silently compile: be honest ... did this never happen to you?

TingPing:
> be honest ... did this never happen to you?

No, probably for a decade or more both `gcc` and `clang` warn if you do this.

The codebase uses the other style and I would prefer consistency.

TingPing:
And I was talking about the *commit* message. The Github issue title is unrelated.

TingPing:
Thank you!

PR #2676, opened 2022-01-26
Replace calls to ctime() with a safer wrapper.

SadieCat:
This wrapper can handle a much bigger date range on 64-bit platforms and will always return a value instead of NULL on error avoiding a potential crash if the server sends a weird value.

Currently for unrepresentable values it returns "?" (where it would have previously returned NULL and potentially crashed) but I'm open to changing this if a better value can be proposed.

### DIFF
--- a/src/common/util.c
+++ b/src/common/util.c
@@ -1307,6 +1307,49 @@ safe_strcpy (char *dest, const char *src, int bytes_left)
 	}
 }
 
+const char *
+safe_timestr(time_t ts)
+{
+	struct tm *timeinfo;
+	static char timebuf[64];
+
+	errno = 0;
+	timeinfo = localtime (&ts);
+	if (!timeinfo)
+	{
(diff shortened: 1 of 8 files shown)


src/common/util.c
```diff
@@ -1307,6 +1307,49 @@ safe_strcpy (char *dest, const char *src, int bytes_left)
 	}
 }
 
+const char *
+safe_timestr(time_t ts)
```
TingPing, line 1311: Would this be more portable and simple?:

```c
GDateTime *datetime = g_date_time_new_from_unix_utc (ts);
GDateTime *localtime = g_date_time_to_local (datetime);
char *timestr = g_date_time_format (datetime, "...");
g_date_time_unref (datetime);
g_date_time_unref (localtime);
return timestr;
```

It results in allocations but thats fine.

TingPing: Docs: https://developer-old.gnome.org/glib/stable/glib-GDateTime.html

PR #2695, opened 2022-04-18
Add AnonOps in the server list

saberan0n:
Hi, I added AnonOps (irc.anonops.com) in the default list, I also had to add `ignoreinvalid` since not supported by default (and AnonOps use a self-signed certificate).
Have a nice day!


TingPing:
I will not add a network that has self-signed certs.

Issue #2699, opened 2022-04-28
Add options in the installer for taskbar/desktop shortcuts


Felipefpl:
It'd be a nice thing to allow us to choose if in the end of the installation we'd like to create options to:

Add a shortcut on desktop/taskbar/start menu or then add none, it's a small nice feature that is surely missed.

TingPing:
A script can easily move the window to a specific display. I'm not sure it really makes sense as an option.

DjLegolas:
@TingPing @Felipefpl - from what I have seen in the InnoSetup script, there are several points to this:
* the start manu is being created when installing in non-portable mode
* the desktop icon and quick launch are not being created at all
* adding shortcuts to quick launch folder, does not add it to the start bar
what do you think is the best solution for this?
is it a good-enough behavior?

Felipefpl:
The best solution would be the ability to create a shortcut to the 3 places: quick launch/desktop/taskbar, desktop is the main option of course since most of apps create shortcuts for it, ;)

DjLegolas:
It took too long to complete... 🤦‍♂️ 
MS removed in windows 10 the ability of programs to pin themselves to taskbar...

Anyway, can you check the MR @Felipefpl? Try the installer from the artifacts.

Felipefpl:
Thank you for the additions but where are the artifacts? I searched github/hexchat's site and i didnt find anything except the common installer.

DjLegolas:
You can find the installer artifacts in the Windows build test:
https://github.com/hexchat/hexchat/actions/runs/2316542167

Felipefpl:
Thank you for the link, i downloaded, installed it and it worked except it didnt create a shortcut on taskbar, i'm on win7 x64 and i got the x86 version.

DjLegolas:
You have in there 2 installers, one for x86 and one for x64. Make sure you have downloaded the correct one :upside_down_face:
Regarding pinning to taskbar, I'm using a third party utility that should work on win7...
But @TingPing don't want to use it anyway so I'll have to find another way to do it.
Maybe in a later MR, as pinning to taskbar is the very easy.

Felipefpl:
I dont understand why using a 32 bit installer would cause problems, never had this problem with other programs before.

I tried 64 bit installer and it didnt work but this time i clicked on details while installing and i noticed the installer didnt download syspin. That must be the reason that the taskbar's shortcut isnt installed.

PR #2708, opened 2022-05-12
add shortcut options in setup

DjLegolas:
will allow the user to control on which of the shortcuts to create:
* start menu - will now show the relevant window
* desktop
* quick launch
* pin to taskbar - added a program for this (syspin)
    https://www.technosys.net/products/utils/pintotaskbar

on uninstall, all will be removed on uninstall, including `syspin.exe`

closes https://github.com/hexchat/hexchat/issues/2699

### DIFF
--- a/win32/installer/hexchat.iss.tt
+++ b/win32/installer/hexchat.iss.tt
@@ -25,7 +25,7 @@ DefaultDirName={pf64}\HexChat
 DefaultDirName={pf32}\HexChat
 #endif
 DefaultGroupName=HexChat
-DisableProgramGroupPage=yes
+AllowNoIcons=yes
 SolidCompression=yes
 Compression=lzma2/ultra64
 SourceDir=..\rel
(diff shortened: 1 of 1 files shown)


win32/installer/hexchat.iss.tt
```diff
@@ -309,6 +323,9 @@ begin
 #endif
 		DOTNET := 'https://dl.hexchat.net/misc/dotnet_40.exe';
 		SPELL := 'https://dl.hexchat.net/hexchat/HexChat%20Spelling%20Dictionaries%20r2.exe';
+		SYSPIN := 'https://www.technosys.net/download.aspx?file=syspin.exe';
```
TingPing, line 326: I don't want to add a third party executable to our installer unless I really trust it. (Which I do for python.org and microsoft).

TingPing:
Outside of that I want to be clear that all of these options should default off.

win32/installer/hexchat.iss.tt
DjLegolas: Can't argue here.
The only issue is finding another way to do this. From what I have found so far, microsoft removed the easy way to do this programmatically...
I'll try to find another way.
However, if you want, it's possible to merge everything except the pinning to taskbar and try to work on it separately.

DjLegolas:
Ok, will be changed it a later rebase.

win32/installer/hexchat.iss.tt
TingPing: Yes lets keep that feature for another day. Its not super valuable IMO.

DjLegolas: Removed.

DjLegolas:
Sorry for taking so long... Here is an updated installer without `Pin To Taskbar` ability.

Felipefpl:
Thanks for fixing this.

Issue #2726, opened 2022-08-21
fe-gtk: notification_plugin_deinit type mismatch


NexAdn:
The function `int notification_plugin_deinit()` has two differing function signatures,
resulting in a type mismatch.

Version 1 (header file): https://github.com/hexchat/hexchat/blob/d7c6c424e8c078d3e5a28c830ff3f6f62da8b39a/src/fe-gtk/plugin-notification.h#L23

Version 2 (source file): https://github.com/hexchat/hexchat/blob/d7c6c424e8c078d3e5a28c830ff3f6f62da8b39a/src/fe-gtk/plugin-notification.c#L248-L250

Original build output: https://861458.bugs.gentoo.org/attachment.cgi?id=794948

Since the function does not use a hypothetical function argument, it's probably safe to use `int(void)` has function signature.

**To reproduce locally**, just build with `CFLAGS="-flto -Werror=lto-type-mismatch"` and `meson -Dplugin=true`

PR #2727, opened 2022-08-21
notification_plugin_deinit: Set function signature to int(void)

NexAdn:
Previously the function signature was inconsistent throughout
src/fe-gtk/plugin-notification.{h,c}: One file had the signature
int(void), while the other had int(void*). Since this type mismatch
might lead to problems (especially with LTO) and the (possibly provided)
function argument isn't used in the function's definition, this commit
sets int(void) as function signature for both the declaration and
definition of the function.

Fixes: https://github.com/hexchat/hexchat/issues/2726


TingPing:
`void*` is the right one.

NexAdn:
PR updated to use int(void *) as signature.

PR #2729, opened 2022-08-28
  Fix PROTOCTL NAMESX and only send when not using `multi-prefix`.

SadieCat:
This capability is the equivalent of the old protoctl token so there's no need to send the old method if the capability is enabled.

This pull request also fixes `PROTOCTL NAMESX` on servers that support protoctl but not capability negotiation.


SadieCat:
Oops got the two legacy extensions mixed up. Fixed.

Issue #2508, opened 2020-10-01
Multiple spell check dictionaries for Windows seems broken


fredddii:
I am unable to get spell checking to work for an additional dictionary: nl_NL

In the installer I have selected spell check dictionaries to be installed (you could not specify any languages however).

In C:\Users\User\AppData\Local\enchant there are only two 0 KB en_GB files: en_GB.dic and en_GB.exc

In my settings I configured: en_GB, nl_NL

Only en_GB is getting spell checked.

eggsgent:
I have the same issue with nb_NO i 2.16.1

TingPing:
HexChat doesn't ship dictionaries on modern Windows, they are provided by Windows itself.

PR #2731, opened 2022-09-01
Support whitespace between language codes

kissen:
So far, when configuring multiple spell check languages, Hexchat requires users to separate multiple entries with commas and only commas. This patch allows users to also enter whitespace, e.g.

    de_DE, en_US

as is common in many applications.

I believe this patch may resolve issue https://github.com/hexchat/hexchat/issues/2508. I know the current behavior of Hexchat was very confusing to me.

### DIFF
--- a/src/fe-gtk/sexy-spell-entry.c
+++ b/src/fe-gtk/sexy-spell-entry.c
@@ -1255,7 +1255,7 @@ void
 sexy_spell_entry_activate_default_languages(SexySpellEntry *entry)
 {
 	GSList *enchant_langs;
-	char *lang, *langs;
+	char *lang, **i, **langs;
 
 	if (!have_enchant)
 		return;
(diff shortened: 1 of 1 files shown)


TingPing:
Can you test this:

`de_DE , en_US`

I suspect `en_US` will not be picked up. This bit of code uses `strtok()` which reads over the string and `g_strstrip()` modifies the string in-place.

I think the ideal fix for that is using `g_strsplit()` instead of `strtok()`.

kissen:
I tested `foo_FAKE, de_DE , en_US` and both `de_DE` and `en_US` are picked up correctly.

That is because `strtok` splits up `foo_FAKE, de_DE , en_US` into `foo_FAKE`, ` de_DE ` and ` en_US`, that is `strtok` inserts terminating zeroes in-place into the string it is passed. These tokens still contain whitespace which is then removed with `g_strstrip`.

`g_strstrip` just chains `g_strchug` and `g_strchomp`, both of which modify the string in-place, but always return the pointer they were passed to. So something like ` de_DE ` = `{' ', 'd', 'e', '_', 'D', 'E', ' ', 0 }` gets turned to ` de_DE ` = `{'d', 'e', '_', 'D', 'E', 0, 0, 0 }`.

That said, the [documentation for `g_strstrip(str)`](https://docs.gtk.org/glib/func.strstrip.html) does not grantee that it always returns `str`. I will re-write this patch to use `g_strsplit()` instead at next convenience.

TingPing:
> I will re-write this patch to use g_strsplit() instead at next convenience.

Thanks! The less `strtok()` is used the better :)

kissen:
I have updated the pull request to use `g_strsplit_set` rather than `strtok`. We now split on comma, spaces and tab characters. I think for most purposes that is enough.

PR #2734, opened 2022-09-18
update servlist.c

deepend-tildeclub:
Added irc.newnet.net to the server list


TingPing:
Honestly I don't think a network with 20-30 users is noteworthy enough to be in the default list.

PR #2737, opened 2022-09-24
Default /SERVER and friends to use SSL when built with OpenSSL.

SadieCat:
Since commit 747a52aae8806a9072a23ea68212767f352ac431 users have to opt-out of using SSL when creating a new server. This commit makes it so /SERVER also uses SSL by default.

In order to connect insecurely users must now use one of these methods:

    /SERVER -insecure irc.example.com
    /SERVER irc.example.com -6667

The `-ssl` flag and the `+port` syntax have been retained for compat reasons.

---

This might be a controversial change but I think its worth doing in order to avoid users accidentally connecting insecurely without realising.

### DIFF
--- a/src/common/outbound.c
+++ b/src/common/outbound.c
@@ -3249,7 +3249,7 @@ cmd_reconnect (struct session *sess, char *tbuf, char *word[], char *word_eol[])
 		int offset = 0;
 
 #ifdef USE_OPENSSL
-		int use_ssl = FALSE;
+		int use_ssl = TRUE;
 		int use_ssl_noverify = FALSE;
 		if (g_strcmp0 (word[2], "-ssl") == 0)
 		{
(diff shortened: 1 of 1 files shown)


src/common/outbound.c
```diff
@@ -4098,14 +4117,14 @@ const struct commands xc_cmds[] = {
...
-	 N_("SERVCHAN [-ssl|-ssl-noverify] <host> <port> <channel>, connects and joins a channel")},
+	 N_("SERVCHAN [-insecure|-ssl|-ssl-noverify] <host> <port> <channel>, connects and joins a channel")},
 #else
 	{"SERVCHAN", cmd_servchan, 0, 0, 1,
 	 N_("SERVCHAN <host> <port> <channel>, connects and joins a channel")},
 #endif
 #ifdef USE_OPENSSL
 	{"SERVER", cmd_server, 0, 0, 1,
-	 N_("SERVER [-ssl|-ssl-noverify] <host> [<port>] [<password>], connects to a server, the default port is 6667 for normal connections, and 6697 for ssl connections")},
+	 N_("SERVER [-insecure|-ssl|-ssl-noverify] <host> [<port>] [<password>], connects to a server, the default port is 6667 for insecure connections, and 6697 for ssl connections")},
```
TingPing, line 4127: Could we better elaborate here SSL is the default

SadieCat: Just pushed a change that I think makes it more clear. How about that?

Issue #2434, opened 2020-02-09
checksum plugin causes freezes


nahemoth33:
After downloading a file, hexchat is freezing, while checksum-plugin is calculating the checksum.

hexchat version: 2.14.3
checksum version: 3.1
OS: debian gnu/linux (testing)

TingPing:
I asked for a reproducer, step by step to get that result including a file example.

nahemoth33:
what is a reproducer? what file-example?
sorry but I don't know what you mean

I only get files sent and after finishing, it freezes, till the checksum is calculated.

nahemoth33:
Hello TingPing,
were you able to determine the bug? or do you need some further information?

TingPing:
Yes the bug is actually quite clear I just have been very busy.

Totto16:
Any Updates on this, since it still occurs?

TingPing:
@Totto16 This function needs to be ran using Gio's Async APIs (or in a thread): https://github.com/hexchat/hexchat/blob/master/plugins/checksum/checksum.c#L128

Totto16:
Ok thanks I'll look into it

Totto16:
This is solved in the commits [here](https://github.com/hexchat/hexchat/commit/a01d9ea1528a33dfc8706b391324ef50afc97991) resulting from PR #2775 
So it can be closed now @TingPing

PR #2743, opened 2022-10-24
Ported the checksum plugin to the async file read functions

Totto16:
Summary: Ported the checksum plugin to the async file read functions and bumped the version up to 3.3:

solves #2434

The checksum plugin now uses the async file read functions, to make non blocking reads.

To achive this in c, since c doesn't support lambdas like e.g c++ every single function that involved something synchronus had to be rewritten completely:

Here is an example pattern for that:

```c
boolean do_something_normal(char* somethings, char *out_buf){

char* local_buffer[BUF_CAP];

int result = do_something_normal(char* some_options);
if(result){
  hexchat_printf(
            ph,
            "some message",
  }
return result;
}
```

```c
void do_something_async(char* somethings, char *out_buf,  void fn_pt(boolean, void *)){

char* local_buffer = malloc[BUF_SIZE];
// handle malloc error

struct xy = malloc(sizeof(struct xy));
// handle malloc error

do_something_async(char* some_options, cb, data_struct);
}

void cb(boolean result, void* data_struct){

if(result){
  hexchat_printf(
            ph,
            "some message",

    do_something_with(data_struct->checksum);

   data_struct->cb(TRUE, data_struct->checksum)
return;
  }
  data_struct->cb(FALSE, data_struct->checksum)
  free(everything);
}
```

Essentially every variable has to be passed around all functions by pointers in malloced memory, that only gets freed, after everything is done with that memory.

This structs are like bound variables in c++, since if we would allocate them in the function do_something_async it would be out of lifetime when we actually need it.

I tested everything multiple times (only in Linux!) in different scenarios, from big (28 GB) to small files. I also ran it with valgrind and some other analytic c tools to check, that I haven't forgotten any frees. But nevertheless, I would be happy if someone would test this too.

### DIFF
--- a/plugins/checksum/checksum.c
+++ b/plugins/checksum/checksum.c
@@ -22,244 +22,431 @@
 
 #include "config.h"
 
-#include <stdlib.h>
-#include <glib.h>
-#include <glib/gstdio.h>
 #include <gio/gio.h>
+#include <glib.h>
+#include <glib/gerror.h>
+#include <glib/gstdio.h>
+#include <stdlib.h>
 
 #include "hexchat-plugin.h"
 
 #define BUFSIZE 32768
-#define DEFAULT_LIMIT 256									/* default size is 256 MiB */
+#define DEFAULT_LIMIT 256 /* default size is 256 MiB */
 #define SHA256_DIGEST_LENGTH 32
 #define SHA256_BUFFER_LENGTH 65
 
-static hexchat_plugin *ph;									/* plugin handle */
+static hexchat_plugin *ph; /* plugin handle */
 static char name[] = "Checksum";
 static char desc[] = "Calculate checksum for DCC file transfers";
-static char version[] = "3.1";
+static char version[] = "3.3";
 
-static void
-set_limit (char *size)
-{
-	int limit = atoi (size);
+static void set_limit(char *size) {
+  int limit = atoi(size);
 
-	if (limit > 0 && limit < INT_MAX)
-	{
-		if (hexchat_pluginpref_set_int (ph, "limit", limit))
-			hexchat_printf (ph, "Checksum: File size limit has successfully been set to: %d MiB\n", limit);
-		else
-			hexchat_printf (ph, "Checksum: File access error while saving!\n");
-	}
-	else
-	{
-		hexchat_printf (ph, "Checksum: Invalid input!\n");
-	}
+  if (limit > 0 && limit < INT_MAX) {
+    if (hexchat_pluginpref_set_int(ph, "limit", limit))
+      hexchat_printf(
+          ph,
+          "Checksum: File size limit has successfully been set to: %d MiB\n",
+          limit);
+    else
+      hexchat_printf(ph, "Checksum: File access error while saving!\n");
+  } else {
+    hexchat_printf(ph, "Checksum: Invalid input!\n");
+  }
 }
 
-static int
-get_limit (void)
-{
-	int size = hexchat_pluginpref_get_int (ph, "limit");
+static int get_limit(void) {
+  int size = hexchat_pluginpref_get_int(ph, "limit");
 
-	if (size <= 0 || size >= INT_MAX)
-		return DEFAULT_LIMIT;
-	else
-		return size;
+  if (size <= 0 || size >= INT_MAX)
+    return DEFAULT_LIMIT;
+  else
+    return size;
 }
 
-static gboolean
-check_limit (GFile *file)
-{
-	GFileInfo *file_info;
-	goffset file_size;
+static gboolean check_limit(GFile *file) {
+  GFileInfo *file_info;
+  goffset file_size;
 
-	file_info = g_file_query_info (file, G_FILE_ATTRIBUTE_STANDARD_SIZE, G_FILE_QUERY_INFO_NONE,
-									NULL, NULL);
+  file_info = g_file_query_info(file, G_FILE_ATTRIBUTE_STANDARD_SIZE,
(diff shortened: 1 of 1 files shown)


TingPing:
You reformatted the entire file, that makes it much harder to review.

plugins/checksum/checksum.c
```diff
@@ -22,244 +22,431 @@
...
+
+  char *word_1 = malloc(strlen(word[1]) + 1);
+  if (word_1 == NULL) {
+    hexchat_printf(ph, "Checksum: Failed to malloc needed memory\n");
+    free(checksum);
+    free(user_data);
+    return HEXCHAT_EAT_NONE;
+  }
+
+  char *word_2 = malloc(strlen(word[2]) + 1);
```
TingPing, line 389: Replace all `malloc` in this file.

Use `g_strdup()` to copy strings. `g_new()` to allocate in general.

Totto16: ok, and sorry, I didn't notice the file was formatted, but autoformatters format automatically I guess ;(

plugins/checksum/checksum.c
```diff
@@ -22,244 +22,431 @@
...
-static gboolean
-check_limit (GFile *file)
-{
-	GFileInfo *file_info;
-	goffset file_size;
+static gboolean check_limit(GFile *file) {
+  GFileInfo *file_info;
+  goffset file_size;
+
+  file_info = g_file_query_info(file, G_FILE_ATTRIBUTE_STANDARD_SIZE,
```
TingPing, line 72: This is a blocking IO call. There is an async version.

TBH I think it may make sense to use `GTask` and run the blocking IO in a thread with `g_task_run_in_thread()`. (Do be careful to not call `hexchat_` API in that thread).

plugins/checksum/checksum.c
```diff
@@ -22,244 +22,431 @@
...
+  if (!file) {
+    hexchat_printf(ph, "Checksum: Failed to open %s\n", filename);
+    return result_cb(FALSE, bound_data);
+  }
+
+  if (!check_limit(file)) {
+    hexchat_printf(ph,
+                   "Checksum: %s is larger than size limit. You can increase "
+                   "it with /CHECKSUM SET.\n",
+                   filename);
```
TingPing, line 239: Since this is no longer blocking, I think we should remove this setting entirely.

It doesn't really matter too much how big it is since we do chunked reads.

plugins/checksum/checksum.c
Totto16: So should I apply these fixes, or reqriet it with the task, I personally think, the async read is better, so maybe run it async in a new thread, or would that be overkill ?

TingPing: No it doesn't make sense to combine the two.

The async way can be fine, its just more code thats harder to read.

TingPing:
Here is what the threaded approach looks like: https://github.com/hexchat/hexchat/tree/wip/checksum/threading

It is untested, has a FIXME and a TODO, but should be close to good.

plugins/checksum/checksum.c
Totto16: true

Totto16:
This also seems file, the FIXME could be resolved, but that would be complicated, since it has to find another context, taht isn't closed yet, so I don't think thats necessary, if you close the context you requested /Sent a file, then no output, that isn't something unbearable behavior.

So the main question here is, which version is "better":

The pros of the task solution is, that it isn't that complicated. But a disadvantage is the synchronous IO, this isn't a problem perse, but it is blocking IO, which is typically slower than async IO, since it has to block and wait additionally, since the data is written in in 32 K blocks, I am not sure how much of a difference this makes to asynchronous IO, I would have to measure that. But generally speaking I have learned and seen, that asynchronous stuff is faster, since the os (at least Linux) has some scheduling, that favors these kinds of tasks, I am not totally sure if that's the case everywhere, but I would favor asynchronous IO. I personally don't know much / nothing of the Glib implementation of async read, so I'm not sure about that. 

So I would prefer the asynchronous approach personally, and I find it "better", since code difficulty isn't seen /noticed by the end user. But you have to decide which version would be better to ship, since both solutions work and I would say in the end it doesn't make that much of a difference. But nevertheless, just tell me what you're liking more, so that we can merge your / my solution :)

TingPing:
I do not believe your performance expectations are correct.

The async flow works like this:

1. `read()` is called
2. if error is WOULD_BLOCK a GSource is created and added to the mainloop
3. the mainloop repeatedly polls the GSource to see if its ready to be read if so go to 1 if not repeat 3

The threaded approach works like this:

1. `read()` is called and does nothing until ready

Totto16:
Like I said, I don't know how Glib implemented the async read :sweat_smile:
So if the Task Solution is better, then let's use that version.

TingPing:
I don't really use this so while I think my approach is good it needs some testing and final cleanups.

Could you take my branch, clean up the fixmes, and test it?

Totto16:
Yes,I I'll do that and look into it :+1: 

Some notes, I am using my version since the discussion started, and had no issues at all (I am using hexchat extensively to download  huge files (Linux ISOs :smile:) onto HDD's )

But since I am also fine and it also is a better idea to use the thread solution I'll fix the remaining FIXMEs and issues and test it extensively :+1:

TingPing:
> Yes,I I'll do that and look into it +1

Thank you!

I don't believe there was any technical issues with your patch. I only believe the threaded version will be a bit more simple and efficient. I really do appreciate the contribution.

Totto16:
I had to close this, since my fork only forked the `main` branch, the continuation of this is   #2775

PR #2760, opened 2023-01-28
Migrate to GTK 3: remove deprecated/removed functions

SadieCat:
This is part one of my attempt to port HexChat to GTK 3. It's a pretty big changeset so I have split up my commits to make reviewing easier. This initial pull request just contains the low hanging fruit which can be dealt with whilst still depending on GTK 2.

---

The following functions which were removed/deprecated in GTK+3 have been replaced with their non-deprecated equivalents:

- `gdk_beep` -> `gdk_desktop_beep`
- `get_drawable_get_size` -> `gdk_window_get_width`, `gtk_window_get_height`
- `gtk_widget_is_composited` -> `gtk_screen_is_composited`

The following function have been removed:

- `gtk_tree_view_set_rules_hint` &mdash; this does nothing on GTK3 and also apparently does nothing with a lot of modern GTK2 themes that I have tried.

### DIFF
--- a/src/fe-gtk/chanlist.c
+++ b/src/fe-gtk/chanlist.c
@@ -459,7 +459,7 @@ chanlist_join (GtkWidget * wid, server *serv)
 			g_snprintf (tbuf, sizeof (tbuf), "join %s", chan);
 			handle_command (serv->server_session, tbuf, FALSE);
 		} else
-			gdk_beep ();
+			gdk_display_beep (gdk_display_get_default ());
 		g_free (chan);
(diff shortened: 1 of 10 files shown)


TingPing:
Looks reasonable so far. How would you like to handle this work?

We could make a new `gtk3` branch and merge as we go into there. That allows us to do transitionary commits that aren't ready yet while not breaking `master`.

SadieCat:
I don't really have a preference. Whatever is most convenient for you.

PR #2772, opened 2023-04-02
Not all servers accept SSL. Therefore, check the box "Use SSL for all…

fabzzap:
… the servers on this network" _only_ when the ssl flag is true for that network


TingPing:
This is the default option. Users can uncheck it if they wish to use insecure networks.

PR #2774, opened 2023-04-03
python: fix for timers that unhook themselves

Celelibi:
When a timer hook unhooks itself and return False, the internal code generates an exception about a weakly-referenced object that no longer exists.

I am unsure how it all works. My tests work, but they might not be exhaustive enough.

In particular, I don't understand why setting `hook.is_unload = True` is necessary. But my guess is that this line can be safely ignored when the `Hook` object has already been destroyed, since the only effect it could have had was in `Hook.__del__`. Which won't be called again because the object no longer exists.

I also note that the code relying on `__del__` seems quite fragile. As far as I understand, the python plugin uses weak references to avoid references loops. This lets the CPython garbage collector free the object as soon as the last reference is destroyed, and thus call `__del__` at that time. As far as I know, this behavior is an [implementation detail](https://docs.python.org/3/reference/datamodel.html#object.__del__) and a more straightforward logic might be warranted in the future.

### DIFF
--- a/plugins/python/python.py
+++ b/plugins/python/python.py
@@ -291,7 +291,15 @@ def _on_timer_hook(userdata):
     if hook.callback(hook.userdata) == True:
         return 1
 
-    hook.is_unload = True  # Don't unhook
+    try:
+        # Avoid calling hexchat_unhook twice if unnecessary
+        hook.is_unload = True
+    except ReferenceError:
+        # hook is a weak reference, it might have been destroyed by the callback
+        # in which case it has already been removed from hook.plugin.hooks and
+        # we wouldn't be able to test it with h == hook anyway.
+        return 0
+
     for h in hook.plugin.hooks:
         if h == hook:
             hook.plugin.hooks.remove(h)


plugins/python/python.py
```diff
@@ -284,13 +284,19 @@ def _on_server_attrs_hook(word, word_eol, attrs, userdata):
...
     if hook.callback(hook.userdata) == True:
         return 1
 
-    hook.is_unload = True  # Don't unhook
-    for h in hook.plugin.hooks:
+    try:
+        hook.is_unload = True  # Don't unhook
+    except ReferenceError:
+        # hook is a weak reference, it might have been destroyed by now
+        pass
```
TingPing, line 295: Would it work to hoist out `hook.is_unload` above `hook.callback` just like you did for plugin?

TingPing:
> In particular, I don't understand why setting hook.is_unload = True is necessary. 

Unload hooks are the only "fake" hook. As-in hexchat doesn't know about them and doesn't have callbacks for them. The python plugin manages them itself.

plugins/python/python.py
TingPing: Do keep the comment that `hook` can be destroyed in `hook.callback` though.

Celelibi:
> > In particular, I don't understand why setting hook.is_unload = True is necessary.
> 
> Unload hooks are the only "fake" hook. As-in hexchat doesn't know about them and doesn't have callbacks for them. The python plugin manages them itself.
> 
> EDIT: Actually I now understand your confusion since this isn't an unload hook. Maybe the variable should be renamed but I don't remember offhand why this is done.

I think I get it now. Correct me if I'm wrong, but I think it's to avoid calling `hexchat_unhook` twice on the same hook.

If we have to cancel the timer (because it returned False), we want to remove the `Hook` object from the `Plugin`. But don't want to explicitly call `hexchat_unhook` on it (as `Hook.__del__` normally would) because returning 0 from `_on_timer_hook` will already unhook it.
https://github.com/hexchat/hexchat/blob/847e5a14d6d7991682c3add853f7b3cc32f935df/src/common/plugin.c#L751-L755

But is there an actual downside to having an explicit call to `hexchat_unhook` in addition to returning 0 from the callback? (Or, equivalently, having several calls to `hexchat_unhook`.) It looks like this case is already taken care of by not doing anything in `hexchat_unhook` if it's already unhooked.
https://github.com/hexchat/hexchat/blob/847e5a14d6d7991682c3add853f7b3cc32f935df/src/common/plugin.c#L972-L977
And by checking if the callback unhooked itself.
https://github.com/hexchat/hexchat/blob/847e5a14d6d7991682c3add853f7b3cc32f935df/src/common/plugin.c#L744-L749

I can't come up with a scenario where we could end up freeing the C hook object and continue using it. At least, not with python code alone.
So, maybe we could just remove this line altogether?

plugins/python/python.py
Celelibi: Hoisting the assignment to `hook.is_unload` would prevent the finalizer `Hook.__del__` from calling `hexchat_unhook`.
Meaning that python code calling `hexchat.unhook(timer)` would remove the `Hook` from the `Plugin`, getting the `Hook` object destroyed without canceling the timer at the C level. It would get called and produce a `ReferenceError` when trying to call the registered python callback.

Setting it back to `False` if the callback returns `True` wouldn't eliminate the need for a `try` if the callback both cancels itself and return `True`.

As far as I understand, setting `hook.is_unload = True` only makes sens right before removing the hook from `plugin.hooks`. But please see my comment below as to why I think it might be ok to just remove this line.

TingPing:
It sounds like it can be done safely. If you test it then it seems fine.

Celelibi:
My tests uncovered that in the loop, `h == hook` would fail if `hook` was already deleted while there were other hooks in the set `plugin.hooks`. Therefore the fix was wrong.

Here I propose something simpler: just don't run the loop if `hook` disappeared (with a long comment). I left the assignment to `hook.is_unload` for three reasons:
1. We need a use of the variable `hook` to raise a `ReferenceError`. There's no other way to test if a weak proxy points to a destroyed variable.
2. Let's not call `hexchat_unhook` once more if we can help it.
3. I wasn't happy with enclosing the whole loop in a `try` statement because a `ReferenceError` could come from multiple places while we only want to catch it when it's raised because `hook` is destroyed. In which case, we souldn't be running the loop in the first place.

But again, I think the right way would be to get rid of the weakref + `__del__` logic, that (I think) makes everything more complex that they need to be.

PR #2775, opened 2023-04-03
Ported the checksum plugin to use threads

Totto16:
This is a continuation of #2743 and solves https://github.com/hexchat/hexchat/issues/2434

Some notes on the changes I made after the initial implementation of @TingPing in [here](https://github.com/hexchat/hexchat/issues/2434):

I tested it for wrong frees / use after frees etc. with `valgrind --tool=memcheck --leak-check=full` 
I tested it in all possible scenarios regarding the context (regarding printing)

So it initially tries to get the context for the sender /receiver nickname, if there's no channel open with that nickname it saves the current channel (which is the channel, where you requested /send the file). After the thread is finished, it tries to set the context to that previously saved context. If that context is destroyed, it prints the message to the current channel, you're in. This is the same behavior as every other function, that print viable information to the user (like e.g a start of a DCC REC etc.)

### DIFF
--- a/plugins/checksum/checksum.c
+++ b/plugins/checksum/checksum.c
@@ -166,7 +166,7 @@ dccoffer_cb (char *word[], void *userdata)
 		ctx = hexchat_find_context (ph, NULL, NULL);
 	}
 
-	ChecksumCallbackInfo* callback_data= g_malloc(sizeof(ChecksumCallbackInfo));
+	ChecksumCallbackInfo *callback_data = g_new (ChecksumCallbackInfo, 1);
 	callback_data->ctx = ctx;
 	callback_data->send_message = TRUE;
 
--- a/plugins/checksum/checksum.c
+++ b/plugins/checksum/checksum.c
@@ -134,7 +134,7 @@ dccrecv_cb (char *word[], void *userdata)
 		ctx = hexchat_find_context (ph, NULL, NULL);
 	}
 
-	ChecksumCallbackInfo* callback_data= g_malloc(sizeof(ChecksumCallbackInfo));
+	ChecksumCallbackInfo *callback_data = g_new (ChecksumCallbackInfo, 1);
 	callback_data->ctx = ctx;
 	callback_data->send_message = FALSE;
 	
--- a/plugins/checksum/checksum.c
+++ b/plugins/checksum/checksum.c
@@ -31,49 +31,46 @@ static char name[] = "Checksum";
 static char desc[] = "Calculate checksum for DCC file transfers";
 static char version[] = "4.0";
 
-static void
-print_sha256_result (hexchat_context *ctx, gboolean send_message, const char *checksum, const char *filename, GError *error)
-{
-	/* Context may have been destroyed. */
-	/* FIXME: This could still send the PRIVMSG even if the context was closed. */
-	if (!hexchat_set_context (ph, ctx))
-		return;
 
-	if (error)
-		hexchat_printf (ph, "Failed to create checksum for %s: %s", filename, error->message);
-	else if (send_message)
+typedef struct {
+	hexchat_context *ctx;
+	gboolean send_message;
+} ChecksumCallbackInfo;
+
+
+static void
+print_sha256_result (ChecksumCallbackInfo *info, const char *checksum, const char *filename, GError *error)
+{
+
+	/* Context may have never been set or destroyed.*/
+	if(!info->ctx  || !hexchat_set_context (ph, info->ctx)){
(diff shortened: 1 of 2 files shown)


plugins/checksum/checksum.c
```diff
@@ -31,49 +31,46 @@ static char name[] = "Checksum";
...
+		hexchat_set_context (ph, ctx);
+	}
 
-	g_free (sha256);
-	g_clear_error (&error);
+	if (error){
+		hexchat_printf (ph, "Failed to create checksum for %s: %s\n", filename, error->message);
+	}
+	else if (info->send_message){
+		hexchat_commandf (ph, "quote PRIVMSG %s :SHA-256 checksum for %s (remote): %s", hexchat_get_info (ph, "channel"), filename, checksum);
```
TingPing, line 56: I think your `hexchat_find_context()` above creates the side-effect that it *may* change to a different network's context. This would then send a message to a user on a different network.

plugins/checksum/checksum.c
```diff
@@ -22,216 +22,164 @@
...
-		hexchat_commandf (ph, "quote PRIVMSG %s :SHA-256 checksum for %s (remote): %s", word[2], word[1], checksum);
+	ctx = hexchat_find_context (ph, NULL, word[3]);
+	// if we didn't send the file from the channel, thats the private channel between us and the receiver, use the current one, from which we actually sent it
+	if(!ctx){
+		ctx = hexchat_find_context (ph, NULL, NULL);
 	}
 
-	return HEXCHAT_EAT_NONE;
-}
+	ChecksumCallbackInfo* callback_data= g_malloc(sizeof(ChecksumCallbackInfo));
```
TingPing, line 169:
```suggestion
	ChecksumCallbackInfo *callback_data = g_new (ChecksumCallbackInfo, 1);
```

plugins/checksum/checksum.c
```diff
@@ -22,216 +22,164 @@
...
 	}
 
-	file_stream = g_file_read (file, NULL, NULL);
-	if (!file_stream)
-	{
-		hexchat_printf (ph, "Checksum: Failed to read file %s\n", filename);
-		g_object_unref (file);
-		return FALSE;
-	}
+	ChecksumCallbackInfo* callback_data= g_malloc(sizeof(ChecksumCallbackInfo));
```
TingPing, line 137:
```suggestion
	ChecksumCallbackInfo *callback_data = g_new (ChecksumCallbackInfo, 1);
```

plugins/checksum/checksum.c
```diff
@@ -22,216 +22,164 @@
...
-	{
-		hexchat_printf (ph, "Checksum: %s is larger than size limit. You can increase it with /CHECKSUM SET.\n", filename);
-		g_object_unref (file);
-		return FALSE;
+	/* Print in the privmsg tab of the sender */
+	ctx = hexchat_find_context (ph, NULL, word[3]);
+	// if we didn't request the file from the channel, it's send to us from, use the current one, from which we actually requested it
+	if(!ctx){
+		ctx = hexchat_find_context (ph, NULL, NULL);
 	}
```
TingPing, line 135: Maybe we should just store the network id and username. Then later find the correct context.

This is a bit more complicated but you can copy this function here:
https://github.com/hexchat/hexchat/blob/847e5a14d6d7991682c3add853f7b3cc32f935df/plugins/fishlim/plugin_hexchat.c#L83-L107

TingPing: Hmm, well that function doesn't default to the front context on the network if unfound. So will need some tweaking.

TingPing: For simplicity sake you could just store the server name and username and then use `hexchat_find_context()`. Falling back to `hexchat_find_context (ph, servername, NULL);` if needed.

Totto16: Seems reasonable, I already thought about doing that, but what should it do, when it also can't find the servername, should it default to the current channel, which I thought is fine, but you said it's wrong, so should it print nothing, which seems also reasonable :thinking:

Totto16: I now did it the simple way. It also solves the issue of a potentially sending a PRIVMSG to the wrong place.

It now stores the servername and the channel and uses these later to find a context. The channel and servername can be potentially NULL, if the initial `hexchat_get_info` returns NULL, as far as I understand this isn't possible, if we just got a dccoffer, but if we send one? I just wanted to ask, if thats possible, and how to handle those cases, I would say to return and print nothing when the start channel and/or servername are NULL to begin with.

TingPing:
Merged manually into master. Thanks!

Issue #2793, opened 2023-07-25
Abort connect when IP fails to bind


konsolebox:
Hexchat probably needs to abort when it fails to bind to configured bind hostname otherwise it will use a network that it isn't supposed to use.

I just created a patch for it but I'm not getting the expected error message.  I'm getting "Unknown host. Maybe you misspelled it?" instead.  This happens both when the IP address to bind to is available or not, and I'm not sure where I made a mistake.

```
diff --git a/src/common/network.c b/src/common/network.c
index fcdaf547..a4a4a7de 100644
--- a/src/common/network.c
+++ b/src/common/network.c
@@ -164,13 +164,15 @@ net_connect (netstore * ns, int sok4, int sok6, int *sok_return)
 	return error;
 }
 
-void
+int
 net_bind (netstore * tobindto, int sok4, int sok6)
 {
-	bind (sok4, tobindto->ip6_hostent->ai_addr,
-			tobindto->ip6_hostent->ai_addrlen);
-	bind (sok6, tobindto->ip6_hostent->ai_addr,
-			tobindto->ip6_hostent->ai_addrlen);
+	if (bind (sok4, tobindto->ip6_hostent->ai_addr,
+			tobindto->ip6_hostent->ai_addrlen) == 0)
+		if (bind (sok6, tobindto->ip6_hostent->ai_addr,
+				tobindto->ip6_hostent->ai_addrlen) == 0)
+			return 0;
+	return -1;
 }
 
 void
diff --git a/src/common/network.h b/src/common/network.h
index 8c1c0c79..5c47b261 100644
--- a/src/common/network.h
+++ b/src/common/network.h
@@ -35,7 +35,7 @@ netstore *net_store_new (void);
 void net_store_destroy (netstore *ns);
 int net_connect (netstore *ns, int sok4, int sok6, int *sok_return);
 char *net_resolve (netstore *ns, char *hostname, int port, char **real_host);
-void net_bind (netstore *tobindto, int sok4, int sok6);
+int net_bind (netstore *tobindto, int sok4, int sok6);
 char *net_ip (guint32 addr);
 void net_sockets (int *sok4, int *sok6);
 
diff --git a/src/common/server.c b/src/common/server.c
index e14da237..8d23ea97 100644
--- a/src/common/server.c
+++ b/src/common/server.c
@@ -909,6 +909,11 @@ server_read_child (GIOChannel *source, GIOCondition condition, server *serv)
[...]

PR #2803, opened 2023-09-15
Abort connect when IP fails to bind

konsolebox:
For https://github.com/hexchat/hexchat/issues/2793.


konsolebox:
Please review.

PR #2809, opened 2023-11-01
Added support for SCRAM-SHA-1, SCRAM-SHA-256 and SCRAM-SHA-512

patrick-irc:
This PR adds the following SASL mechanisms: SCRAM-SHA-1, SCRAM-SHA-256, SCRAM-SHA-512. I also moved PLAIN to an own function and fixed a compile error when USE_OPENSSL is false.

TODO for hexchat developers:
* Add localized error messages in src/common/inbound.c line 1982 and 2009
* Update documentation

If you are looking for a server to test on, you can use nova.irc.it and register for SASL at https://accounts.contempt.chat/
After adding a credential, please wait a few minutes for synchronization.

RFCs:
* https://datatracker.ietf.org/doc/html/rfc5802
* https://datatracker.ietf.org/doc/html/rfc7677


src/common/inbound.c
```diff
@@ -1766,6 +1775,24 @@ get_supported_mech (server *serv, const char *list)
...
+			if (!strcmp(mechs[i], "SCRAM-SHA-256")) {
+				ret = MECH_SCRAM_SHA_256;
+				break;
+			}
+		}
+		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_512) {
+			if (!strcmp(mechs[i], "SCRAM-SHA-512")) {
+				ret = MECH_SCRAM_SHA_512;
+				break;
+			}
```
TingPing, line 1794:
```suggestion
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_1)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-1"))
			{
				ret = MECH_SCRAM_SHA_1;
				break;
			}
		}
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_256)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-256"))
			{
				ret = MECH_SCRAM_SHA_256;
				break;
			}
		}
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_512)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-512"))
			{
				ret = MECH_SCRAM_SHA_512;
				break;
			}
```

src/common/inbound.c
```diff
@@ -1766,6 +1775,24 @@ get_supported_mech (server *serv, const char *list)
...
+		}
```
TingPing, line 1795:
```suggestion
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_1)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-1"))
			{
				ret = MECH_SCRAM_SHA_1;
				break;
			}
		}
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_256)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-256"))
			{
				ret = MECH_SCRAM_SHA_256;
				break;
			}
		}
		else if (serv->loginmethod == LOGIN_SASL_SCRAM_SHA_512)
		{
			if (!strcmp(mechs[i], "SCRAM-SHA-512"))
			{
				ret = MECH_SCRAM_SHA_512;
				break;
			}
        }
```

src/common/inbound.c
```diff
@@ -1901,6 +1932,90 @@ inbound_cap_list (server *serv, char *nick, char *extensions,
 								  NULL, NULL, 0, tags_data->timestamp);
 }
 
+static void plain_authenticate(server *serv, char *user, char *password)
+{
```
TingPing, line 1936:
```suggestion
static void
plain_authenticate(server *serv, char *user, char *password)
{
```

src/common/inbound.c
```diff
@@ -1901,6 +1932,90 @@ inbound_cap_list (server *serv, char *nick, char *extensions,
...
+	if (pass_len % 400 == 0)
+		tcp_sendf (serv, "AUTHENTICATE +\r\n");
+}
+
+#ifdef USE_OPENSSL
+/*
+ * Sends AUTHENTICATE messages to log in via SCRAM.
+ */
+static void scram_authenticate(server *serv, const char *data, const char *digest,
+							   const char *user, const char *password)
```
TingPing, line 1971:
```suggestion
static void
scram_authenticate(server *serv, const char *data, const char *digest,
				   const char *user, const char *password)
```

src/common/inbound.c
TingPing: I won't comment everywhere, but if you could reformat to this style.

src/common/inbound.c
TingPing: This is the style also.

src/common/inbound.c
```diff
@@ -1901,6 +1932,90 @@ inbound_cap_list (server *serv, char *nick, char *extensions,
...
+		tcp_sendf(serv, "AUTHENTICATE +\r\n");
+		scram_free_session(serv->scram_session);
+		serv->scram_session = NULL;
+	} else if (ret == SCRAM_ERROR) {
+		// Authentication failed
+		tcp_sendf(serv, "AUTHENTICATE *\r\n");
+
+		if (serv->scram_session->error != NULL) {
+			// TODO: localized output
+			// g_warning("SASL SCRAM authentication failed: %s", serv->scram_session->error);
```
TingPing, line 2010: You could leave this as a `g_info()`, won't be shown by default but helpful for debugging. Same goes for any other prints you left behind if you think they're useful.

src/common/inbound.c
```diff
@@ -1901,6 +1932,90 @@ inbound_cap_list (server *serv, char *nick, char *extensions,
...
+		}
+
+		scram_free_session(serv->scram_session);
+		serv->scram_session = NULL;
```
TingPing, line 2014:
```suggestion
        g_clear_pointer (&serv->scram_session, scram_free_session);
```

src/common/inbound.c
TingPing: Also the style is spaces after functions `foo (arg);`

src/common/inbound.c
```diff
@@ -1921,50 +2036,38 @@ inbound_sasl_authenticate (server *serv, char *data)
...
 
 void
 inbound_sasl_error (server *serv)
 {
+#ifdef USE_OPENSSL
+	if(serv->scram_session != NULL)
+	{
+		scram_free_session(serv->scram_session);
+		serv->scram_session = NULL;
+	}
```
TingPing, line 2069:
```suggestion
    g_clear_pointer (&serv->scram_session, scram_free_session);
```

src/common/scram.c
```diff
@@ -0,0 +1,297 @@
...
+
+	// The server can append his nonce to the client's nonce
+	if (strlen(server_nonce_b64) < client_nonce_len ||
+	    strncmp(server_nonce_b64, session->client_nonce_b64, client_nonce_len)) {
+		session->error = g_strdup_printf("Invalid server nonce: %s", server_nonce_b64);
+		return SCRAM_ERROR;
+	}
+
+	salt_len = g_base64_decode_step((gchar *) salt, strlen((char *) salt), (guchar *) salt,
+	                                &b64_state, &b64_save);
```
TingPing, line 179: Everywhere I see you use `g_base64_decode_step()` it seems like `g_base64_decode_inplace()` would be more appropriate?

src/common/scram.c
```diff
@@ -0,0 +1,297 @@
...
+
+	// SaltedPassword := Hi(Normalize(password), salt, i)
+	session->salted_password = malloc(session->digest_size);
```
TingPing, line 182:
```suggestion
	session->salted_password = g_malloc (session->digest_size);

```

src/common/scram.c
```diff
@@ -0,0 +1,297 @@
...
+		return SCRAM_ERROR;
+	}
+}
+
+int scram_process(scram_session *session, const char *input, char **output, size_t *output_len)
+{
+	int ret = 0;
+
+	switch (session->step) {
+	case 0:
```
TingPing, line 279: It would be nice to turn the steps into an enum with readable names.

src/common/server.c
```diff
@@ -1765,7 +1765,13 @@ server_set_defaults (server *serv)
 	g_free (serv->chanmodes);
 	g_free (serv->nick_prefixes);
 	g_free (serv->nick_modes);
-
+#ifdef USE_OPENSSL
+	if(serv->scram_session != NULL)
+	{
+		scram_free_session (serv->scram_session);
+		serv->scram_session = NULL;
+	}
```
TingPing, line 1773:
```suggestion
        g_clear_pointer (&serv->scram_session, scram_free_session);
```

src/common/server.c
```diff
@@ -1937,6 +1943,12 @@ server_free (server *serv)
 #ifdef USE_OPENSSL
 	if (serv->ctx)
 		_SSL_context_free (serv->ctx);
+
+	if(serv->scram_session != NULL)
+	{
+		scram_free_session (serv->scram_session);
+		serv->scram_session = NULL;
+	}
```
TingPing, line 1951:
```suggestion
        g_clear_pointer (&serv->scram_session, scram_free_session);
```

src/common/scram.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef HEXCHAT_SCRAM_H
+#define HEXCHAT_SCRAM_H
+
+#include "config.h"
+#ifdef USE_OPENSSL
+#include <openssl/evp.h>
+
+#define SCRAM_ERROR 0
+#define SCRAM_IN_PROGRESS 1
+#define SCRAM_SUCCESS 2
```
TingPing, line 27: enum might be nice.

TingPing (changes requested):
Overall this looks excellent. Mostly tiny nit-picks.

The Visual Studio project files also need to be updated. Shouldn't be too hard to edit them by hand to add the files.

src/common/scram.h
patrick-irc: Added enum scram_status.

src/common/scram.c
patrick-irc: This is more a step counter, so int makes sense imho. Also it is only evaluated in scram_process().

0 = nothing done so far, send client-first-message
1 = parse server-first-message and send client-final-message
2 = parse/verify server-final-message

src/common/inbound.c
patrick-irc: Changed to g_info()

patrick-irc:
Thank you for your fast reply. The code is ready for another review :)

inbound.c:1992
```
// TODO: localized output
g_warning ("Could not create SCRAM session with digest %s", digest);
```

I don't think that this error will ever occur, but if so, it must be visible to the user in hexchat (not only in console or logs) - as there is no message from the ircd. How should I output it and do we need to localize it? I still have not figured out how localization is working.

Thanks.

TingPing:
If its printed in the UI it should be translated.

To do so:

```c
#include <glib/gi18n.h>

// Anywhere a string is used wrap it with the `_()` function.
char* example = _("Text");
```

Then add the file to `po/POTFILES`.

src/common/scram.c
TingPing: Ok, not a big deal.

TingPing:
I believe a simple `PrintTextf()` is probably how to print some of these unlikely issues.

src/common/scram.c
```diff
@@ -0,0 +1,323 @@
...
+	client_final_message_without_proof = g_strdup_printf ("c=biws,r=%s", server_nonce_b64);
+
+	session->auth_message = g_strdup_printf ("%s,%s,%s", session->client_first_message_bare,
+											 data, client_final_message_without_proof);
+
+	// ClientKey := HMAC(SaltedPassword, "Client Key")
+	client_key = g_malloc0 (session->digest_size);
+
+	HMAC (session->digest, session->salted_password, session->digest_size,
+		  (unsigned char *) CLIENT_KEY, strlen (CLIENT_KEY), client_key, &client_key_len);
```
TingPing, line 217: I'm not an expert, but surely key being a known value isn't ideal?

TingPing: Well, never mind:

> where the [XOR](https://en.wikipedia.org/wiki/XOR) operation is applied to byte strings of the same length, H(ClientKey) is a normal hash of ClientKey. 'Client Key' and 'Server Key' are verbatim strings.

TingPing:
Windows CI is failing, hopefully this fixes it and you can rebase: https://github.com/hexchat/hexchat/pull/2810

patrick-irc:
> I believe a simple `PrintTextf()` is probably how to print some of these unlikely issues.

Ok, done for EN/DE/IT.

src/common/inbound.c
```diff
@@ -1901,11 +1938,102 @@ inbound_cap_list (server *serv, char *nick, char *extensions,
...
+		g_clear_pointer (&serv->scram_session, scram_free_session);
+	}
+	else if (status == SCRAM_ERROR)
+	{
+		// Authentication failed
+		tcp_sendf (serv, "AUTHENTICATE *\r\n");
+
+		if (serv->scram_session->error != NULL)
+		{
+			g_info ("SASL SCRAM authentication failed: %s", serv->scram_session->error);
```
TingPing, line 2024: Should this be printed to the user?

patrick-irc: Yes, it makes sense, because it reveals errors which are not caused by invalid credentials (very rare). Output added.

TingPing:
Merged as 9b76b557ecaece2a5fa862ea4dc75ed613e3fbf0

Thanks!

